PR #50, opened 2021-12-16
Versions changes

nirarg:
1. Use capi version v1beta1 (changed from v1alpha4)
2. Use Go version 1.17 (changed from 1.16)


config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api/capk-manager:dev
+      - image: localhost:49293/capk-manager-amd64:dev
```
rmohr, line 11: This change seems to be not intentional.

go.mod
```diff
@@ -1,30 +1,81 @@
...
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	k8s.io/apiextensions-apiserver v0.23.0-alpha.4 // indirect
+	k8s.io/kube-openapi v0.0.0-20210817084001-7fbd8d59e5b8 // indirect
+	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b // indirect
+	kubevirt.io/containerized-data-importer-api v1.41.0 // indirect
+	kubevirt.io/controller-lifecycle-operator-sdk v0.2.1 // indirect
+	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
```
rmohr, line 74: There are a lot of replaces. What exactly is the reason for this? Is it the v1beta1 api from capi?

rmohr (changes requested):
Look pretty good overall. A few comments inline.

go.mod
nirarg: The section you pointed on is "require" section
This is auto created after the update to go 1.17
There are 2 `require` sections which are auto generated 
There is one `replace` section which include only 3 lines

config/default/manager_image_patch.yaml
nirarg: You right, removed this change

go.mod
rmohr: > The section you pointed on is "require" section

Indeed :+1:

davidvossel (approved):
/lgtm

I don't have lgtm permissions yet, but take that as an offer of my support :D 

The code looks good to me and func tests passed for me locally. We'll need an OWNER to issue the `/ok-to-test` command in order for CI to execute the e2e tests

rmohr:
@nirarg  I need to fix something in the integration test. I have to bump it to 1.17 too: 

```
../../../../pkg/mod/k8s.io/client-go@v0.23.0-alpha.4/plugin/pkg/client/auth/exec/metrics.go:21:2: package io/fs is not in GOROOT (/root/.gimme/versions/go1.15.8.linux.amd64/src/io/fs)
```

nirarg:
> @nirarg I need to fix something in the integration test. I have to bump it to 1.17 too:

I rebased the PR on this fix and changed Golang version in the Dockerfile

nirarg:
/hold
Merge decreasing to v1alpha1 version in different PR:
https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/pull/54

nirarg:
/hold cancel
@cchengleo I rebased this PR after #54 was merged
Can you please test this too?

cchengleo (approved):
I also verified this PR in our CI and passed. Thanks for the great job!

PR #53, opened 2021-12-17
Make integration tests independent of go version changes

rmohr:

With the new Dockerfile in place with #46  which builds the manager binary, the
binary does not have to be built upfront with the external golang
version. Instead the internal one can be used which makes for instance
integration jobs independent of the go versions installed there.

**Release notes**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,7 +1,7 @@
 on: [push, pull_request]
-name: Test
+name: build
 jobs:
-  test:
+  unit_test:
     strategy:
       matrix:
         go-version: [1.16.x, 1.17.x]
--- a/kubevirtci
+++ b/kubevirtci
@@ -65,7 +65,6 @@ function kubevirtci::down() {
 
 function kubevirtci::build() {
 	export REGISTRY="localhost:$(cluster-up/cluster-up/cli.sh ports registry)"
-	make manager
 	make docker-build
 	make docker-push
 }


rmohr:
/cc @cchengleo 


/cc @nirarg 

@nirarg once this PR is in, you also need to bump the golang version in the Dockerfile in PR #50: https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/main/Dockerfile#L19

We are referencing the golang versions still in too  many places, but we will get to a more sane model :)

davidvossel (commented):
/lgtm

yep, we don't need the `make manager` now that `make docker-build` uses the multi stage docker build. Thanks

cchengleo:
> @nirarg once this PR is in, you also need to bump the golang version in the Dockerfile in PR #50: https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/main/Dockerfile#L19
> 
> We are referencing the golang versions still in too many places, but we will get to a more sane model :)

I think later we should unify the default golang version for the repo.

cchengleo (approved):

nirarg:
> @nirarg once this PR is in, you also need to bump the golang version in the Dockerfile in PR #50: https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/main/Dockerfile#L19

Thanks @rmohr
I changed the version in my PR

PR #67, opened 2022-01-13
Add the new --fail-with-job flag to let mkpj fail when the prowjob fails

rmohr:

https://github.com/kubernetes/test-infra/pull/24686 landed a while ago, the new mkpj flag can be used now.
The github action will now also fail when the prowjob failed and not just indicate when the prowjob is done.

The change will only be effective after this PR with the change is merged.

**Release notes**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -20,4 +20,4 @@ jobs:
           KUBECONFIG: ${{secrets.KUBECONFIG}}
       - name: Test
         run: |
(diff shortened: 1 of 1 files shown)


nunnatsa:
I'm not sure about the `pull_request_target` trigger. But the condition in the action works in other places in when using `pull_request`. It prevent the action to run on your fork, but it does run on the upstream repo upon PR.

rmohr:
> I'm not sure about the `pull_request_target` trigger. But the condition in the action works in other places when using `pull_request`. It prevent the action to run on your fork, but it does run on the upstream repo upon PR.

Thanks @nunnatsa you are right. It seems like it just took a wile for the actions to pick up the label change. After re-checking everything looks good.

PR #71, opened 2022-01-18
Run e2e test also on label events

rmohr:

Ensure that we react to a label change on the PR to ensure that the e2e
tests get triggered if someone writes /ok-to-test some time after the PR
got created.

`labeled` is the new event. The other three events are the default values.

Right now, if the label gets added to PRs some time later, github will not reconsider if it has to run the integration test unless the code in the PR gets repushed.

**Release notes**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,4 +1,7 @@
-on: [pull_request_target, push]
+on:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


rmohr:
@agradouski that should fix the reason on why on your PR the integration tests did not start after I added /ok-to-test.

PR #79, opened 2022-01-27
Fix bug in KubevirtCluster Reconciler - endless reconcile loop with external controller

nunnatsa:
When the KubevirtCluster CR is manage by an external controller, sometimes there is an update race between capk and the external controller, where each update in one hand, triggers reconcile and update in the other hand and vice versa.

This PR changes that so in this case, the CR is filtered out from the reconciling watch, and so capk will ignore changes made by an external controller.

**Release notes**:

```release-note
```


davidvossel (approved):
/lgtm

For further context, this is using a standardized annotation that the cluster-api uses to signal that a platform's cluster object should not be reconciled by it's controller, but instead it is externally reconciled by another controller.

Many of the other cluster-api platforms do this exact same thing, and it's necessary for how we're consuming capk. Here's the example of where aws provider does this, https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L318

PR #80, opened 2022-01-31
expose cluster load balancer spec and metadata via KubevirtCluster resource

### LINKED ISSUE #64: Tenant cluster loadbalancer service is hardcoded to ClusterIP type
With ClusterIP type a tenant cluster will only be accessible within the same cluster, and cannot be exposed to external workloads.

Example load balancer service:
```
kvcluster-lb      ClusterIP   10.108.169.113   <none>        6443/TCP    4m56s
```

Currently, there's no way to configure the loadbalancer service, as the spec is hard coded.

agradouski:
This PR enables the way to expose cluster load balancer spec via `KubevirtCluster` object. This allows users to specify different types of service for cluster load balancer, as opposed to it always being hardcoded to `ClusterIP`. `cluster-template` has an example of how to specify a service type, but it's optional. By default, `ClusterIP` service type is used.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #64

**Release notes**:

```release-note
NONE
```


api/v1alpha1/kubevirtcluster_types.go
```diff
@@ -85,6 +88,16 @@ type SSHKeys struct {
...
 }
 
+// LoadBalancerServiceTemplate describes the template for the cluster load balancer service.
+type LoadBalancerServiceTemplate struct {
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +nullable
+	// Service metadata.
+	ObjectMeta metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Service specification.
+	Spec corev1.ServiceSpec `json:"spec,omitempty" valid:"required"`
```
rmohr, line 98: Do you think it would make sense to introduce here a ServiceSpecTemplate? Otherwise it is pretty hard to understand which fields are considered. It seems to only be the `Type` right now. Right? API validation would let pass all kinds of fields which are finally ignored.

rmohr (changes requested):
One recommendation.

api/v1alpha1/kubevirtcluster_types.go
agradouski: yes that makes sense. originally, I thought we could expose the entire service spec to users (and it theory that could also work, we could set selector and ports in the template, too), but at the same time we might not want to complicate `KubevirtCluster` spec too much, so, for now I just exposed the service type, and service metadata, with potential to expose other service spec parts in future if needed. 

but agree, for now, we can just have our own ServiceSpecTemplate, to keep things explicit 👍

agradouski: also, note, that service spec is optional in KubevirtCluster object, and by default we fall back to the ClusterIP type as before. the main use case for now is being able to specify a different type (e.g. LoadBalancer) as well as setting labels and annotations for the service.

agradouski:
@rmohr addressed your recommendation, please take another look

api/v1alpha1/kubevirtcluster_types.go
```diff
@@ -85,6 +91,26 @@ type SSHKeys struct {
 	DataSecretName *string `json:"dataSecretName,omitempty"`
 }
 
+// LoadBalancerServiceTemplate describes the template for the cluster load balancer service.
+type LoadBalancerServiceTemplate struct {
```
rmohr, line 95: It may make sense to explicitly state what exactly gets routed over this service. Like: apiserver traffic, ingress, ...

api/v1alpha1/kubevirtcluster_types.go
```diff
@@ -43,10 +43,16 @@ type KubevirtClusterSpec struct {
...
+	// +optional
 	SshKeys SSHKeys `json:"sshKeys,omitempty"`
 
 	// InfraClusterSecretRef is a reference to a secret with a kubeconfig for external cluster used for infra.
+	// +optional
 	InfraClusterSecretRef *corev1.ObjectReference `json:"infraClusterSecretRef,omitempty"`
+
+	// LoadBalancerServiceTemplate describes load balancer service template.
+	// +optional
+	LoadBalancerServiceTemplate LoadBalancerServiceTemplate `json:"loadBalancerServiceTemplate,omitempty"`
```
rmohr, line 55: It may make sense to explicitly state what exactly gets routed over this service. Like: apiserver traffic, ingress, ...

maybe it makes sense to not call this `LodaBalancer...` since in fact we can also choose non-load-balancer service types and instead use a name which hints what exactly is routed over this service.

rmohr (commented):
One comment which may help users understanding what they exactly configure with this new spec. The implementation itself looks good to me!

rmohr:
here a retriggered e2e test which is green: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubernetes-sigs_cluster-api-provider-kubevirt/80/pull-kubernetes-sigs-cluster-api-provider-kubevirt-e2e/1488464929697566720

We merged https://github.com/kubevirt/project-infra/pull/1885 in kubevirts CI system which hopefully fix the GOPROXY issues from now on.

api/v1alpha1/kubevirtcluster_types.go
agradouski: agree, will update. historically, we called it a "load balancer" in a sense of it being a load balancer for control plane nodes, even though as far as implementation goes it's a simple service. maybe make more sense to name it "control plane service" in the spec to be generic, and add more details in the api docs as you suggest 👍

agradouski:
@rmohr addressed, please take another look

rmohr (approved):
/lgtm

Looks great!

PR #119, opened 2022-03-28
Add missing go compiler to the e2e test

nunnatsa:

```release-note
None
```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -10,6 +10,10 @@ jobs:
     steps:
       - name: checkout code
         uses: actions/checkout@v3
+      - name: Install Go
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -12,6 +12,10 @@ jobs:
         uses: actions/checkout@v3
       - name: compile clusterkubevirtadm
         run: make clusterkubevirtadm-linux
+      - name: Install Go
```
rmohr, line 15: does `make clusterkuevirtadm-linux` need a compiler? If so, the order probably has to be changed.

nunnatsa: sure. fixing

PR #129, opened 2022-04-05
Change "kubevirtci create-cluster" command to use clusterkubevirtadm kubeconfig

nirarg:

This commit changes kubevirtci to behave like the tenant cluster is created on ext infra
This is done in order to test the cluster creation with clusterkubevirtadm limited kubeconfig
This will also allow us to follow another resources would be added in the future to the tenant cluster creation
Notice: added special behavior for kubevirtci, in which the "localhost" IP in the kubeconfig is replaced
with the controlplane VM IP, in order to allow using the kubeconfig inside the cluster

### DIFF
--- a/kubevirtci
+++ b/kubevirtci
@@ -96,11 +96,19 @@ function kubevirtci::install() {
 	${_kubectl} wait -n capk-system --for=condition=Available=true deployment/capk-controller-manager --timeout=10m
 }
 
+function kubevirtci::generate_kubeconfig() {
+        make clusterkubevirtadm-linux
+        bin/clusterkubevirtadm-linux-amd64 apply credentials --namespace e2e-test
+        bin/clusterkubevirtadm-linux-amd64 get kubeconfig --namespace=e2e-test --output-kubeconfig=kubeconfig-e2e
+        sed -i -r 's/127.0.0.1:[0-9]+/192.168.66.101:6443/g' kubeconfig-e2e
+}
+
 function kubevirtci::create_cluster() {
 	export NODE_VM_IMAGE_TEMPLATE=quay.io/kubevirtci/fedora-kubeadm:35
 	export IMAGE_REPO=k8s.gcr.io
 	export CRI_PATH="/var/run/crio/crio.sock"
-	$CLUSTERCTL_PATH generate cluster kvcluster --kubernetes-version v1.21.0 --control-plane-machine-count=1 --worker-machine-count=1 --from templates/cluster-template.yaml | ${_kubectl} apply -f -
+	oc create secret generic external-infra-kubeconfig -n capk-system --from-file=kubeconfig=kubeconfig-e2e --from-literal=namespace=e2e-test
+	$CLUSTERCTL_PATH generate cluster kvcluster --kubernetes-version v1.21.0 --control-plane-machine-count=1 --worker-machine-count=1 --from templates/cluster-template-ext-infra.yaml | ${_kubectl} apply -f -
 }
 
 kubevirtci::fetch_kubevirtci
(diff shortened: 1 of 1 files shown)


kubevirtci
```diff
@@ -95,11 +95,19 @@ function kubevirtci::install() {
 	${_kubectl} wait -n capk-system --for=condition=Available=true deployment/capk-controller-manager --timeout=10m
 }
 
+function kubevirtci::generate_kubeconfig() {
+        make clusterkubevirtadm-linux
+        bin/clusterkubevirtadm-linux-amd64 apply credentials --namespace e2e-test
+        bin/clusterkubevirtadm-linux-amd64 get kubeconfig --namespace=e2e-test --output-kubeconfig=kubeconfig-e2e
+        sed -i -r 's/127.0.0.1:[0-9]+/192.168.66.101:6443/g' kubeconfig-e2e
```
agradouski, line 102: please clarify `192.168.66.101`

nirarg: The kubevirtci environment has special hook to access this cluster, it change the IP to localhost and use port forward in order to access into this special environment
This hook works when you access the cluster from the hosting machine
This PR creates different requirement, the kubeconfig also has to be used from the cluster itself, from capk manager container, in this case we can't use the localhost IP (because we are not running in the hosting machine)
Therefore I changed the IP to the control plane (master) node IP (which is always `192.168.66.101`), this IP also can be accessed from the capk manager container

agradouski: 👍

kubevirtci
```diff
@@ -96,11 +96,19 @@ function kubevirtci::install() {
 	${_kubectl} wait -n capk-system --for=condition=Available=true deployment/capk-controller-manager --timeout=10m
 }
 
+function kubevirtci::generate_kubeconfig() {
+        make clusterkubevirtadm-linux
+        bin/clusterkubevirtadm-linux-amd64 apply credentials --namespace e2e-test
+        bin/clusterkubevirtadm-linux-amd64 get kubeconfig --namespace=e2e-test --output-kubeconfig=kubeconfig-e2e
```
nunnatsa, line 102: Do we want the namespace to be hard coded? Maybe it can be the default, but that makes the function not usable for any other purpose than e2e test.

kubevirtci
```diff
@@ -96,11 +96,19 @@ function kubevirtci::install() {
...
+        sed -i -r 's/127.0.0.1:[0-9]+/192.168.66.101:6443/g' kubeconfig-e2e
+}
+
 function kubevirtci::create_cluster() {
 	export NODE_VM_IMAGE_TEMPLATE=quay.io/kubevirtci/fedora-kubeadm:35
 	export IMAGE_REPO=k8s.gcr.io
 	export CRI_PATH="/var/run/crio/crio.sock"
-	$CLUSTERCTL_PATH generate cluster kvcluster --kubernetes-version v1.21.0 --control-plane-machine-count=1 --worker-machine-count=1 --from templates/cluster-template.yaml | ${_kubectl} apply -f -
+	oc create secret generic external-infra-kubeconfig -n capk-system --from-file=kubeconfig=kubeconfig-e2e --from-literal=namespace=e2e-test
+	$CLUSTERCTL_PATH generate cluster kvcluster --kubernetes-version v1.21.0 --control-plane-machine-count=1 --worker-machine-count=1 --from templates/cluster-template-ext-infra.yaml | ${_kubectl} apply -f -
```
nunnatsa, line 111: We need to think about moving to K8s 1.23; of course, in another PR.

nunnatsa (commented):
Added 2 inline questions.

kubevirtci
nirarg: I don't think that the namespace name is important, because the `clusterkubevirtadm` taking care to create it if it doesn't exists
Currently this function purpose is only for e2e-test, In case we would like to use it to another purpose in the future, we would the option to modify it

kubevirtci
nirarg: Agree, but different PR

nunnatsa (approved):

PR #145, opened 2022-06-30
Graceful deletion of a VirtualMachineInstance

nunnatsa:
When a host node is deleted, the guest VMIs are evicted, killing the
running tenant node in this VMI.

This PR gracefully delete the VMI, by first drain the guest node. Only
if the drain successful, CAPK will delete the VMI.

The PR is based on a feature in KubeVirt, where by setting the eviction
strategy to "external" KubeVirt will not delete the VMI, but will set
the `vmi.Status.EvacuationNodeName` field to signal to the external
controller (CAPK in this case) that the VMI should be evicted.

## Known Issues
* does not support external clusters (#100)

**Release notes**:

```release-note
Added graceful eviction of a VMI
```


davidvossel (commented):
great start!

I left a few comments in line and have one kind of higher level comment.

We need a drain timeout, where if a drain doesn't complete after `x` minutes (regardless of how many times drain is retried), we delete the VMI regardless. This timeout is meant to ensure that we won't block the removal of a VMI indefinitely when a infra node is being drained.

controllers/vmi_controller.go
```diff
@@ -0,0 +1,212 @@
+package controllers
```
nirarg, line 1: I think need to give more meaningful name to this file, something like 'vmievacuation_controller.go

controllers/vmi_controller.go
```diff
@@ -0,0 +1,212 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	cluster, err := r.getCluster(ctx, vmi)
+	if err != nil {
+		logger.Error(err, "Can't get the cluster form the VirtualMachineInstance")
+		return ctrl.Result{}, err
+	}
+
+	nodeDrained, retryDuration, err := r.drainNode(ctx, cluster, nodeName, logger)
```
nirarg, line 67: Why there should be cases returning `nodeDrained=false` instead of just returning the error?

controllers/vmi_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil || !nodeDrained {
+		// logging done in the drainNode method
+		return ctrl.Result{RequeueAfter: retryDuration}, err
+	}
+
+	// now, when the node is drained, we can safely delete the VMI
+	propagationPolicy := metav1.DeletePropagationForeground
+	err = r.Delete(ctx, vmi, &client.DeleteOptions{PropagationPolicy: &propagationPolicy})
```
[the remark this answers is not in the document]
nirarg: Agree, can be validated even before calling to `drainNode`

controllers/vmi_controller.go
nunnatsa: > Why there should be cases returning `nodeDrained=false` instead of just returning the error?

Because for most cases (but not all of them), we don't want to retry. returning error with the result will cause retry. The calling function can't know which error to return with the result and which error to ignore.

controllers/vmi_eviction_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	if err = kubedrain.RunCordonOrUncordon(drainer, node, true); err != nil {
+		// Machine will be re-reconciled after a cordon failure.
+		logger.Error(err, "Cordon failed")
+		return false, 0, errors.Errorf("unable to cordon node %s: %v", nodeName, err)
+	}
+
+	if err = kubedrain.RunNodeDrain(drainer, node.Name); err != nil {
+		// Machine will be re-reconciled after a drain failure.
+		logger.Error(err, "Drain failed, retry in 20s", "node name", nodeName)
+		return false, 20 * time.Second, nil
```
davidvossel, line 196: we need a global timeout for drain as well. Something that ensures we eventually give up attempting to drain and delete the VMI after `x` minutes. 10 minutes would likely be a pretty conservative number to use.

davidvossel (commented):
great stuff, i left one comment in line.

Have you thought about how to e2e test this?

controllers/vmi_eviction_controller.go
davidvossel: an annotation on the VMI could be used to track when the drain begins. The timeout could be anchored to the start time recorded in the annotatino.

e2e/create-cluster_test.go
```diff
@@ -788,5 +791,60 @@ var _ = Describe("CreateCluster", func() {
...
+					GracePeriodSeconds: pointer.Int64(60 * 10), // 10 minutes
+				},
+			})
+		}).WithTimeout(time.Minute*5).
+			WithPolling(time.Second*10).
+			Should(Succeed(), "pod eviction was not done on time")
+
+		Eventually(func() types.UID {
+			vmiAfterEviction, err := virtClient.VirtualMachineInstance(namespace).Get(chosenVMI.Name, &metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
```
davidvossel, line 841: there's goign to be a brief period of time where the old VMI is deleted, and the new VMI doesn't exist. So this might return an error during that time. 

Maybe just ignore the error if the error indicates the VMI doesn't exist?

davidvossel (commented):
@nunnatsa It occurred to me that the current `IsTerminal()` logic in the machine controller [1] probably needs to be udpated with your pr.

Previously I had made some changes that set the failureReason on the kv machine when a terminal (non recoverable) situation was encountered. One such situation is when the EvactuationNodeName is set when a VMI is not migratable. 

With your change here, we should be able to handle that situation gracefully as long as the RunStrategy == Always. 

1. https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/41c77a793ccbcaa21a05d9b76fd2d4f02c4934f5/pkg/kubevirt/machine.go#L99

controllers/vmi_eviction_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	nodeDrained, retryDuration, err := r.drainNode(ctx, cluster, vmi.Status.EvacuationNodeName, logger)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: retryDuration}, err
+	}
+
+	if !nodeDrained && r.waitingForTimeout(ctx, vmi, logger) {
+		return ctrl.Result{RequeueAfter: retryDuration}, nil
+	}
```
davidvossel, line 93: The drain can block indefinitely due to the way this logic is structured. The timeout needs to be checked either regardless if the drainNode() function returns an error or before drainNode is even called. If the global timeout occurs, the VMI should be deleted.

We don't want anything within the guest cluster to indefinitely block an infra cluster node from being drained.

e2e/create-cluster_test.go
```diff
@@ -844,3 +920,65 @@ var _ = Describe("CreateCluster", func() {
...
+func getRecreatedVMI(vmiName string, namespace string, originalUID types.UID) *kubevirtv1.VirtualMachineInstance {
+	var (
+		vmi *kubevirtv1.VirtualMachineInstance
+		err error
+	)
+	Eventually(func(g Gomega) types.UID {
+		vmi, err = virtClient.VirtualMachineInstance(namespace).Get(vmiName, &metav1.GetOptions{})
+		g.Expect(err).ShouldNot(HaveOccurred())
+		g.Expect(vmi).ShouldNot(BeNil())
+
```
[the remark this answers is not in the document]
nunnatsa: This is why this function uses the  `g Ginkgo` parameter. Using just `Expect` will fail the test, but using `g.Expect` will fail only this cycle.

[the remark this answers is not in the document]
davidvossel: ah interesting to know, thanks!

PR #160, opened 2022-07-25
Select kubeconfig current context namespace before 'default'.

### LINKED ISSUE #158: Tenant namespace should fallback to current context namespace if external kubeconfig is specified.

Discussed with @agradouski:

The namespace where to spawn a tenant cluster should be resolved as follow:

1. the namespace specified by the enduser in a node's VirtualMachine template; i.e. as `KubevirtMachineTemplate.spec.template.spec.virtualMachineTemplate.metadata.namespace`
1. If this is not set and an `infraClusterSecretRef` is set for this cluster: 
   1. If set, use the namespace directly set as a value in the secret.
   1. Otherwise, fallback to the namespace of the context currently selected in the `kubeconfig` set in the secret.
   1. If the latter is not set, then fallback to the `default` namespace
1. Otherwise, use the "current" namespace where the cluster specification has been created.

The current logic skips `2.ii`; i.e. it ignores the namespace, if any, set in the `kubeconfig` file for the current context and directly fallback to `default`.

If the end-user did not specified a namespace in the virtual-machine template, and they reference a secret with a `kubeconfig` to connect to the infrastructure, but the secret doesn't specified directly a namespace, then the namespace of the context currently selected in the `kubeconfig` should be used, if it is set.

**Environment:**

- KubeVirt version: **v1**

/kind bug

pjaton:

When the end-user is requesting a tenant cluster to be created in an external infrastructure cluster, the logic that resolves the namespace where to spawn it doesn't take the namespace set for the current context in the `kubeconfig` that has been provided (please see the issue for the full resolution logic).

This change addresses this by leveraging the `clientcmd.ClientConfig`; which offers a `Namespace` method that will return the namespace of the current context, if it is set, and fallback to `"default"` otherwise.

fixes #158

This fix simply replace the direct call to `clientcmd.RESTConfigFromKubeConfig` to generate the REST config, with the intermediate call to first generate the `clientcmd.ClientConfig`, capture the namespace out of it, and then, the REST config. The intermediate generation of the `clientcmd.ClientConfig` is encapsulated in the `clientcmd.RESTConfigFromKubeConfig` method, we just do it manually here.

**Release notes**:

```release-note
NONE
```

### DIFF
--- a/pkg/infracluster/infracluster.go
+++ b/pkg/infracluster/infracluster.go
@@ -43,20 +43,25 @@ func (w *infraCluster) GenerateInfraClusterClient(infraClusterSecretRef *corev1.
 		return nil, "", errors.New("Failed to retrieve infra kubeconfig from secret: 'kubeconfig' key is missing.")
 	}
 
-	namespace := "default"
-	namespaceBytes, ok := infraKubeconfigSecret.Data["namespace"]
-	if ok {
+	clientConfig, err := clientcmd.NewClientConfigFromBytes(kubeConfig)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "failed to create K8s-API client config")
+	}
+
+	namespace, _, err := clientConfig.Namespace()
+	if err != nil {
+		return nil, "", errors.Wrap(err, "failed to resolve namespace from client config")
+	}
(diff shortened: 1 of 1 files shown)


davidvossel (approved):
/approve

PR #167, opened 2022-08-01
Fix infra secret lookup so that it defaults to the namespace of the referencing resource

### LINKED ISSUE #165: CAPK should look for `infraClusterSecret` in the `KubevirtCluster` resource namespace by default

Request the creation of a cluster in an external infrastructure providing the kubeconfig in a secret in the same namespace where the cluster spec resources are created and omitting the `Namespace` field in the `KubevirtCluster.spec.infraClusterSecretRef`.

E.g. with a ``KubevirtCluster` like the following one

```yaml
kind: KubevirtCluster
metadata:
    name: pja-multi
spec:
    infraClusterSecretRef:
        apiVersion: v1
        kind: Secret
        name: external-infra-kubeconfig
```

CAPK controller-manager will fail to reconcile the creation of the cluster and report in the log that it cannot find the `external-infra-kubeconfig`.

The `Namespace` field is optional in all the resource references used by CAPI. If the namespace is not specified, then the default behavior is to look for the referenced resource in the same namespace and the referring one. We should have the same logic here.

This is due to the logic in [/pkg/infracluster/infracluster.go #GenerateInfraClusterClient](https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/a8097c471d38aabe38179cd187a4d95618a834d2/pkg/infracluster/infracluster.go#L36) that does not check whether the `Namespace` field is set or not. It the latter case, it should fallback and looked in the provided `ownerNamespace`.

**Environment:**

- KubeVirt version: **v1**

/kind bug

pjaton:

This change fixes where the `GenerateInfraClusterClient` function looks up for the infrastructure secret referenced in a `KubevirtCluster` or `KubevirtMachine`, when no namespace is specified in the `ObjectReference`.

This corrects the current logic by ensuring that, in this case, the code looks for the secret in the same namespace than the referencing resource; which aligns with the default behavior that CAPI follows for other `ObjectReference`s.

To validate this change of behavior, this PR also introduce a set of tests for the `infracluster` package.

fixes #165

Note that this change is split in two commits:

* The first one with a refactoring that extract the duplicated `setupScheme()` function into as single function in the `testing` package.
* The second commit introduces the fix and the tests validating the `infracluster` package.

**Release notes**:

```release-note
NONE
```

### DIFF
--- a/controllers/kubevirtcluster_controller_test.go
+++ b/controllers/kubevirtcluster_controller_test.go
@@ -7,10 +7,7 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	kubevirtv1 "kubevirt.io/api/core/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	. "sigs.k8s.io/controller-runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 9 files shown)


davidvossel (commented):
/approve

PR #185, opened 2022-09-27
spike: Set worker networkings secondary

qinqon:
Hammer the infra to have one controller one worker tenant cluster using secondary interfaces + bridge cni + nmstate

TODO: 
- [x] Basic scenario of one worker harcoding one ip for worker and one for node.
- [x] Basic secenario for two workers with harcoded IPs
- [x] Test live migration during a TCP connection 
- [x] Generate IPs for workers and infra nodes at capk controller
- [x] Set ip from the tenant range on the nodes with NNCP at capk controller
- [x] Use netmask from adquiered IP (right now is harcoded to 24)
- [x] Use different network per tenant
- [x] Do infra node network teardown (delete vlan and bridge).
- [ ] Set iptables masquerade at infra node primary interface from the capk controller.
- [ ] Check services with custom endpoints.
- [ ] Scale up infra cluster (new nncp need to be created)
- [ ] Use secondary networks at control-plane too (kubeinit is called at cloud-init)
- [ ] Check integration at hypershift.
- [ ] Check if we can use proxy ARP so we have the same default gw on tenant nodes
  - https://chandanduttachowdhury.wordpress.com/2017/06/29/using-a-router-for-switching/
  - https://www.sobyte.net/post/2022-06/proxy-arp-calico/

FIXME:
- [ ] Don't modify guest VM networking for live migration (default gw need to be changed to node's IP)
  - Use virt-launcher sidecar pod to do the routing with one IP only accessible  to the VM and repated at all the enodes

**Release notes**:

```release-note
NONE
```


qinqon:
After running the worker VM at difffernt node than controller-plane VM the worker does not join to the cluster.

qinqon:
There is a chicken egg issue, we need to create one NNCP per tenant cluster + node since it will contain the specific ip for that tenant cluster and for that we need to know where the VMI machine is going to run but the VMI depends on the bridge to be created or the virt-launcher pod will be on "Scheduling" forever.

We have to prepare configure all the infra workers nodes where VMIs will run with the bridge and and some IPs from the ipamer, that should be done at KubevirtCluster reconcile.

maiqueb (commented):
Some nits. Looks good. 

I may have a couple more suggestions next week (trying to articulate my thoughts properly ...).

**EDIT:** this is a first ... I wrote this on the **wrong** PR. Please disregard it. (I can't delete the comment 0_o).

PR #187, opened 2022-09-29
ci: Check make generate has being call

qinqon:
Not pushing generated artifacts is a big issue. This change add a check-gen make target and a sigthub action with it.

**Release notes**:

```release-note
NONE
```


davidvossel (commented):
/lgtm
/approve

whoa, good catch! crazy we missed this for so long. thanks!

PR #195, opened 2022-10-12
Fix CRI_PATH var in template files

fengye87:

**Release notes**:

```release-note
Fix CRI_PATH var in template files
```


fengye87:
@agradouski This is an obvious typo in the template file. I found it when I was trying capv and I manually fixed the typo before using the template and it worked. I haven't tried with the original template, it may or may not work, but I think this typo is very obvious.

PR #203, opened 2022-10-28
Spike: guest dhcp companion

qinqon:
Spike to test tenant cluster with two interfaces at nodes but using dnsmasq container to assing IPs (using cloud init is not always possible, ej openshift).

The north/south traffic goes over the pod network the east/west go over the multus stable IP, the VMs are live-migratable.

CHECKS:
- [x] lb service (traffic arrives at secondary pod interface on workers so it's fine)
- [x] worker tenant node advertise the multus IP
- [x] Check kubectl exec on tenant worker pod usin apiserver behind LoadBalancer kubeconfig
- [x] live migration with a tcp connection (ssh or netcat)
  - workers are migratable and kubelet keep connected 
  - ssh using masquerade interface drops connection (as expected)

TODO:
- [x] Use cluster network to configure dnsmasq.
- [x] hypershift need ConfigDrive but we cannot use networkData there, we will need to activate DHCP at the ubuntu image in a different way.
  - If it's external cluster use ConfigDrive if is not use NoCloud with networkData, an alternative is to merge whatever is on the templates
- [x] Assign dnsmasq pods ip address with whereabouts. 
- [x] Generate mac addresses to ip mapping and configure dnsmasq with it, for spike purposes restarting dnsmasq to update config (creationg/deletion kubevirtmachine) should be enough. 
- [x] Change replicas to 2 at dnsmasq to have active-active
- [ ] Check if control-plane works fine with same networking as workers.
  - It ends up with the masquerade address on the "kubernetes" svc endpoints 
```yaml
  subsets:
  - addresses:
    - ip: 10.0.2.2
    ports:
    - name: https
      port: 6443
      protocol: TCP
``` 
- [ ] virtctl port-forward on workers is not working (./kubevirtci ssh-tenant broken)
- [ ] Hack virt-launcher to replace bridge binding with a proxy ARP binding so we can have only one interface at guest with stable default gw too.
- [ ] Keep pod interface as the primary VMI interface but select multus address for kubelet join at workers.
  - https://github.com/kubernetes/kubeadm/issues/203#issuecomment-478206793
  - https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#providing-instance-specific-configuration-details

  

Questions:
- Do we need access the stable IP out of the tenant cluster ? 

readings:
- multiple static Dnsmasq for active-active, https://medium.com/@krisiasty/highly-available-static-dhcp-service-with-dnsmasq-95a9b856b6fb

**Release notes**:

```release-note
```


rmohr:
> Spike to test tenant cluster with two interfaces at nodes but using dnsmasq container to assing IPs (using cloud init is not always possible, ej openshift).

Just curious, is ignition there not sufficient?

PR #209, opened 2022-11-23
release: Add all the cluster template flavors

qinqon:
There were missing template flavors at the github release used when a user generate a cluster with --infrastructure kubevirt, this change add all of the flavors and also check that the kccm ones are generated.

The github action has being refactored for simplicity

**Release notes**:

```release-note
Add all the cluster flavors
```

### DIFF
--- a/.github/workflows/create_release.yaml
+++ b/.github/workflows/create_release.yaml
@@ -44,52 +44,12 @@ jobs:
         run: |
           make create-infrastructure-components
       - name: Create Release
-        id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


nunnatsa (commented):
Overall, looks good. Using the `action-gh-release action` makes the code much simpler and easier to read and maintain.

One question. I think you'll need to add the new created file `templates/cluster-template-lb-kccm.yaml` to git and to this PR (or if it's not needed, then add another filter to drop its creation). WDYT?

nunnatsa:
Now I see that this is the reason that the check-gen lane has failed.

qinqon:
> Now I see that this is the reason that the check-gen lane has failed.

Fixed @nunnatsa can you take another view ?

nunnatsa (approved):

PR #214, opened 2022-11-29
templates: Activate net multiqueue

qinqon:
This change the templates VM to be with to improve performance by using net multiqueue [1] .

[1] https://www.linux-kvm.org/page/Multiqueue#Multiqueue_virtio-net

**Release notes**:

```release-note
Activate network multiqueue at cluster templates.
```


davidvossel (commented):
/approve
/lgtm

Issue #230, opened 2023-03-31
Use an alternative machine bootstrap flag probing strategy (no SSH)


BarthV:
**What steps did you take and what happened:**

At our company, we are building an infrastructure "highly secure" with several contraints imposed by French/European sovereign Cloud Label. Running these constraints made us locate CAPI management cluster in a specific network and managed clusters in other ones.

Working around these rules, we recently tried to block all the traffic between CAPI cluster network and target managed clusters networks (and also disabled SSH daemons in all our VMs).
In order to schedule and manage clusters lifecycle, we expected CAPI/CAPK to only reach managed cluster's apiserver using exposed loadbalanced endpoints, which is open the rest of the network using underlying Kubevirt LB capabilities.

In fact [we discovered (here)](https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/3af474489d6625741111e31465d51bc6a981df68/pkg/kubevirt/machine.go#L227-L240) that CAPK requires a direct SSH access to the VM IP in order to validate CAPI Machine bootstrap success (using [CAPI sentinel file convention](https://cluster-api.sigs.k8s.io/developer/providers/bootstrap.html#sentinel-file)).
Also, this seems to be the unique SSH command I've found in the whole CAPK source code.

At the end with this restriction, CAPK is never able to correctly provision a single kubevirt VM, because VM bootstrap is never acknowledged.

```bash
$ kubectl get machine -n capknossh 
NAME                            CLUSTER   NODENAME   PROVIDERID   PHASE          AGE     VERSION
capknossh-cp-lxpzz              capknossh                         Provisioning   9m41s   v1.26.2
capknossh-wk-6996b7555c-98sgs   capknossh                         Pending        9m41s   v1.26.2
capknossh-wk-6996b7555c-fhm8z   capknossh                         Pending        9m41s   v1.26.2
```

The CAPI specification leaves the infrastructure providers free to choose the sentinel file verification methodology.
So I'd like to start a topic and try finding solutions to avoid doing such SSH connections, which are a very sensitive topic for us.
At the end, I'd love to have a more "read only" & auditable/secure way to check Machine bootstrap status

Possible anwsers could be : 

1.  Add a flag to completely skip SSH bootstrap file check
[...]

PR #231, opened 2023-04-02
Add a new optional http VM bootstrap check method 

BarthV:
**First "really naive" proposal for #230
This is more a basis for discussion than a real proposal that can be merged as it is.**
**It must be discussed, refined and probably implemented in a more elegant way than that.**

This implementation uses go stdlib http client without any tuning, and have added some tests related to checkMode selection. Since TLS-PSK ciphers are not supported yet ( https://github.com/golang/go/issues/6379 ) , I've drop the idea to leverage this kind of authentication / crypto between the controller and the VM for the moment.

It implements a new optional `virtualMachineBootstrapCheck` structure for `KubevirtMachine` & `KubevirtMachineTemplate` Custom Resources.
```yaml
kind: KubevirtMachineTemplate
metadata:
  name: yet-another-machine-template
spec:
  template:
    spec:
      virtualMachineBootstrapCheck: # optional
        mode: http # possible values are : "ssh" or "http" (defaults to ssh, for backward compatibility)
        port: 6440 # only used for http (defaults to 6440)
      virtualMachineTemplate:
        metadata:
          [...]
        spec:
          [...]
```

As an alternative to SSH, a new contract is expected on the VM side: a web server must listen on the chosen port (6440 by default) and expose the `bootstrap-success.complete` Sentinel File to `http://<VM_IP>:<PORT>/bootstrap-success.complete`.

There's currently no http auth (the bootstrap status is public), but with more time a TLS-PSK or mTLS mechanism could be imagined between controller and its managed VM. At this moment, Certs would need to be managed by the controller the same way as SSH keys.

In this patch, SSH keys are still managed, injected and mandatory to the controller global behaviour. But this point could be perfectly enhanced to make SSH related stuff completely optional when Bootstrap Check is made through HTTP. But I don't consider it as a blocker : at the moment, we can perfectly keep ssh management even if bootstrap file check is not done through SSH

Let me know if this makes sense for you or not (and if we should dig deeper in this direction !).
Thanks


BarthV:
Following last Syncup Meeting, we dropping this idea in favor of :
* short term : bypassing Sentinel file check with a "none" policy
* mid term : working on a alternate qemu-guest-agent usage to check file
* long term : making the whole SSH logic optional (for keygen & cloud-init management...) with something like a Sentinel File check mode selector in spec

closing

Issue #233, opened 2023-04-25
Bypass Sentinel file check with a "none" policy


BarthV:
## Context

Following last Syncup Meeting (Tuesday April 18), we are now targeting multiple new features :

1. short term : bypassing Sentinel file check with a "none" policy
2. mid term : working on a alternate qemu-guest-agent usage to check file
3. long term : making the whole SSH logic optional (for keygen & cloud-init management...) with something like a Sentinel File check mode selector in spec

This Issue tracks short term **item 1.**

## Feature Description

Define & implement a Sentinel Check file mode selector.
By default, this selector must preserve current CAPK behavior (SSH check)
We need to add a new "None" mode, which bypass Sentinel file check and act like machine were always bootstrapped.

* SSH keys are still managed and provisionned all the time on the VMs, even if Sentinel file is not checked.
* Configuration should be tested & flexible enough for future enhancements (Actions items 2. & 3. )

## CRD Change Proposal

Adding a new optional string field in KubevirtMachine (& KubevirtMachineTemplate) CR.
This string would describe how [CAPI Sentinel File](https://cluster-api.sigs.k8s.io/developer/providers/bootstrap.html#sentinel-file) is checked for a single Machine.
Since this value is set on the machine, multiple check strategies can live in the same cluster without any problem.
Sanity checks and default value should be implemented to filter values & to make SSH the default option.

Expected value would be : `none` and `ssh`. Empty value will fallback to `ssh`, unexpected values would fails the `KubevirtMachine` reconciliation (or maybe fallback to `ssh` too , but it really can be risky and deceiving for the end user).

**New `KubevirtMachine.spec` field :**
```yaml
virtualMachineBootstrapCheck:
  checkStrategy: [ "none" | "ssh" ]
```

**Example in a template :**
```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
kind: KubevirtMachineTemplate
metadata:
  name: capi-cluster1-cp
  namespace: capi
spec:
  template:
    spec:
      virtualMachineBootstrapCheck:
        checkStrategy: none
      virtualMachineTemplate:
        metadata:
          namespace: capi
        spec:
          runStrategy: Always
          template:
            spec:
              terminationGracePeriodSeconds: 60
              evictionStrategy: External
              [...]
```

BarthV:
`virtualMachineBootstrapCheck` could also be named `virtualMachineBootstrap`. This is probably shorter.

[...]

PR #239, opened 2023-04-26
Add sentinel bootstrap file check modes in CRD

BarthV:
https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/issues/233


agradouski:
should we include this in kubevirtmachine spec explicitly as part of this PR, given that default value is "ssh" anyway:
```
virtualMachineBootstrapCheck:
         checkStrategy: ssh
```
I know we made it backward compatible, and this is not required in the template, asking more wrt potentially confusing existing users.

one advantage of making it explicit is perhaps giving users heads up about changes (current and future) in that part of logic.

@davidvossel @pjaton  wdyt?

pjaton (approved):
Thanks @BarthV , LGTM 👍

pjaton:
> should we include this in kubevirtmachine spec explicitly as part of this PR, given that default value is "ssh" anyway:
> 
> ```
> virtualMachineBootstrapCheck:
>          checkStrategy: ssh
> ```
> 
> I know we made it backward compatible, and this is not required in the template, asking more wrt potentially confusing existing users.
> 
> one advantage of making it explicit is perhaps giving users heads up about changes (current and future) in that part of logic.
> 
> @davidvossel @pjaton wdyt?

In this case, I feel that the expected behavior by most users is really about "please use the default mechanism"; so, using unset to mean default feels more intuitive to me than making this required. I do see your point about future change, though, I simply see that there would be more impact if this is made required.

agradouski (approved):

PR #247, opened 2023-06-30
Frequently Asked Questions

feitnomore:
As documentation is still in very initial phase, I thought it would be nice to have a little FAQ for main questions I've got around the tool when testing it myself.

Thanks

**What this PR does / why we need it**: This just adds a simple FAQ document, so that the project can add documentation in an easy way while the official documentation is not ready.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,0 +1,84 @@
+# Frequently Asked Questions
+
+The present document lists a couple of frequently asked questions from our users and adopters.
+
+## How do I use x CNI with cluster?
+
+There is a great chance your tenant cluster CNI will conflict with infrastructure's cluster CNI.
+To avoid issues, you need first to make sure you are using VXLAN on your CNI, and a customized PORT. Below you can find a few examples:
+
+### Calico
+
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -0,0 +1,84 @@
...
+    {
+      "Network": "10.243.0.0/16",
+      "Backend": {
+        "Type": "vxlan",
+        "Port": 1234
+      }
+    }
+```
+
+## Whenever I reboot a VM in Kubevirt, it looses is IP address, how do I fix it?
```
davidvossel, line 32: looses/loses

also, this shouldn't be a problem. restarting a vm will result in a new ip, but i believe kubernetes is resilient to nodes having a different ip after restart.

does this for sure cause an issue?

feitnomore: image

This is what I was trying to document here.

Thanks

davidvossel: ah yes, ha, i remember that now. I had experimented with worker nodes receiving new IP addresses after restart with some success, but it is true that control plane nodes offer a different set of challenges

davidvossel (approved):
/lgtm
/approve

PR #257, opened 2023-08-17
Bump golang to 1.20

nunnatsa:

**Release notes**:

```release-note
None
```


.github/workflows/clusterkubevirtadm-test.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.18
+        go-version: 1.20
```
qinqon, line 18: Can we retrieve the golang version from `go.mod` with https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file so we don't have to harcode it here.

qinqon (changes requested):

PR #266, opened 2023-09-15
kccm: Bump to v0.5.1

qinqon:
The latest kccm fails if there are no IPs at LB services after some time, this change bump to it and configure the new timeout value.

```release-note
Bump kccm to v0.5.1 it will fail if there is no IPs at LB services after some time.
```


qinqon:
/hold

Let's use this PR for kccm 0.5.1

davidvossel (commented):
/lgtm
/approve

orenc1 (approved):
/lgtm

Issue #274, opened 2023-12-15
Infrastructure resources require Privileged pod security standards


bzub:
**What steps did you take and what happened:**

Install the the provider via:
```bash
clusterctl generate provider --infrastructure="kubevirt"
```

The deployment uses a hostpath volume which requires a privileged pod security standard and requires a docker socket to be found on the host.

**What did you expect to happen:**

Installs resources that meet the `Baseline` or `Restricted` pod security standards.

https://kubernetes.io/docs/concepts/security/pod-security-standards/

**Environment:**

- Cluster-api version: `v1.5.4`
- Cluster-api-provider-kubevirt version: `v0.1.8`
- Kubernetes version: (use `kubectl version`): `v1.28.4`
- KubeVirt version: `v1.1.0`
- OS (e.g. from `/etc/os-release`): Talos

/kind bug
[One or more /area label. See https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/labels?q=area for the list of labels]

bzub:
I removed the docker socket hostpath volume and volumeMount manually from the deployment. It is not needed for a production deployment, it seems.

PR #275, opened 2023-12-15
fix: remove hostpath volume and privileged context

bzub:
This change ensures the manager deployment meets restricted pod security
standards.

Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/issues/274


davidvossel (commented):
/approve

PR #276, opened 2023-12-25
Remove the setup counter from the KVMachine ready status

nunnatsa:
## What this PR does / why we need it
Cluster-API takes the reason and the message from the infra-machine (in our case, the KubeVirtMachine) `Ready` condition, into the cluster-api `Machine`'s `InfrastructureReady` condition. Currently the text is something like `"0 of 2 completed"`. This is not very informative.

This PR changes the behavior so now the KV Machine Ready condition will include both the reason and the message from the failed condition.

That way, the cluster-api `InfrastructureReady` condition, will also contain the same reason and message. This will make the actual issue, if exists, to more accessible to the user.

**Release notes**:
```release-note
Remove the setup counter from the KV-Machine ready status
```


davidvossel (commented):
/lgtm

PR #289, opened 2024-06-06
Allow e2e multiple times

nunnatsa:
One of the e2e tests creates a secret that is not get deleted at the end
of the test. When running the test again, it fails on secret
re-creation.

This commit fixes the issue by deleting the secret at the end of each
test.

**Release notes**:

```release-note
None
```


davidvossel (commented):
/lgtm
/approve

PR #303, opened 2025-01-12
Bump several dependencies to fix vulnerable packages

nunnatsa:

Dependency go:golang.org/x/net:v0.25.0 is vulnerable

Upgrade to 0.33.0

GHSA-w32m-9786-jp63,  Score: 8.7

An attacker can craft an input to the Parse functions that would be processed non-linearly with respect to its length, resulting in extremely slow parsing. This could cause a denial of service.

Read More: https://osv.dev/vulnerability/GHSA-w32m-9786-jp63

GO-2024-3333,  Score: 8.7

An attacker can craft an input to the Parse functions that would be processed non-linearly with respect to its length, resulting in extremely slow parsing. This could cause a denial of service.

Read More: https://osv.dev/vulnerability/GO-2024-3333

**Release notes**:

```release-note
fix vulnerable packages
```


orenc1 (approved):
/lgtm

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/crypto v0.23.0
+	golang.org/x/crypto v0.31.0
```
orenc1, line 13: why not v0.32.0 if already available[1]?

---
[1] https://pkg.go.dev/golang.org/x/crypto

PR #305, opened 2025-01-15
Bump all dependencies

nunnatsa:
**Release notes**:

```release-note
Bump all dependencies
```


oshoval:
Hi, please note that we will soon change Kubevirtci default CNI from Calico to Flannel
you can already try that using latest tag, by setting
`export KUBEVIRT_FLANNEL=true`

Will appreciate if you can bump please and test to see it works without problems,
thank you

PR #343, opened 2025-11-02
Add flags for qps and leaderElection

nikParasyr:
Add new flags to the manager in order to configure:
- QPS related fields for the client
- leader election leaderDuration, renewDeadline and retryPeriod

**Release notes**:

```release-note
```


agradouski:
is there immediate need for this @nikParasyr ? is this consistent with other cluster-api providers ?

PR #355, opened 2026-01-04
Downgrade go version to v1.24.6

nunnatsa:
Some CI tools are still working with go1.24.6 images, and can't build images with later versions of go.

This PE downdrade github.com/google/pprof to the previous version, because it requires go1.24.9, then downgrade go to 1.24.6

**Release notes**:

```release-note
None
```


nunnatsa:
Closing for now. looking for another approach.

PR #361, opened 2026-01-07
Revert moving to claster-api v1beta2

nunnatsa:
Going back to v1beta1, because some user can't move to cluster-api v1.11, and can't adopt v1beta2, yet


qinqon (approved):
/lgtm

PR #382, opened 2026-04-08
fix: clear FailureReason/FailureMessage when VM recovers from terminal state

qinqon:

When a KubeVirt VM is stopped (`virtctl stop` changes `runStrategy` to `Halted`), the VMI reaches a finalized state and `IsTerminal()` correctly sets `FailureReason=UpdateError` on the KubevirtMachine. However, when the VM is started back (`virtctl start` changes `runStrategy` back to `Always`), `IsTerminal()` returns `false` but the `FailureReason`/`FailureMessage` fields are **never cleared**.

This causes the CAPI Machine `status.phase` to stay stuck at `Failed` even though all Machine conditions (`Ready`, `InfrastructureReady`, `NodeHealthy`) correctly show `True`. Downstream consumers that rely on the Machine phase — such as HyperShift's endpointslice reconciler for `default-ingress-passthrough-service` — then permanently mark endpoints as not ready, breaking ingress into hosted clusters.

The fix adds an `else` clause to nil out `FailureReason` and `FailureMessage` when `IsTerminal()` returns `false`, allowing VMs to properly recover from a previously terminal state.

**Which issue this PR fixes**: [OCPBUGS-77929](https://issues.redhat.com/browse/OCPBUGS-77929)

The bug manifests as follows:
1. `virtctl stop VM` → `runStrategy: Halted` → VMI finalizes → `IsTerminal()=true` → `FailureReason` set
2. `virtctl start VM` → `runStrategy: Always` → new VMI created → `IsTerminal()=false` → but `FailureReason` NOT cleared (no `else` clause)
3. `KubevirtMachine.Status.FailureReason` stays set → CAPI Machine `phase: Failed` → stuck permanently

Evidence from the customer environment:
- `KubevirtMachine`: `ready=true`, `failureReason=UpdateError` (contradictory)
- `Machine`: conditions all `True`, `phase=Failed` (contradictory)
- `EndpointSlice`: `ready=false` (downstream breakage)

**Release notes**:
```release-note
Fix KubevirtMachine FailureReason/FailureMessage not being cleared when a VM recovers from a terminal state, which caused the CAPI Machine phase to stay stuck at "Failed" permanently.
```

### DIFF
--- a/controllers/kubevirtmachine_controller.go
+++ b/controllers/kubevirtmachine_controller.go
@@ -261,6 +261,9 @@ func (r *KubevirtMachineReconciler) reconcileNormal(ctx *context.MachineContext)
 		failureErr := capierrors.UpdateMachineError
 		ctx.KubevirtMachine.Status.FailureReason = &failureErr
 		ctx.KubevirtMachine.Status.FailureMessage = &terminalReason
+	} else {
+		ctx.KubevirtMachine.Status.FailureReason = nil
+		ctx.KubevirtMachine.Status.FailureMessage = nil
 	}
 
 	// Provision the underlying VM if not existing
--- a/controllers/kubevirtmachine_controller_test.go
+++ b/controllers/kubevirtmachine_controller_test.go
@@ -34,7 +34,8 @@ import (
 
 	"sigs.k8s.io/cluster-api-provider-kubevirt/pkg/kubevirt"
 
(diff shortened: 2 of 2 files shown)


nunnatsa (approved):
/approve

nunnatsa:
ci passed

/unhold

PR #384, opened 2026-05-24
fix: set ownerReference on bootstrap userdata Secret to VM

amasolov:

Sets an ownerReference from the `-userdata` bootstrap Secret to its VirtualMachine in `reconcileKubevirtBootstrapSecret`. This establishes a Kubernetes-level lifecycle relationship between the two resources, providing cascading cleanup (VM deleted -> Secret garbage-collected) and allowing external controllers to determine the Secret is still in use.

Without this, the `-userdata` Secret can be orphaned if `reconcileDelete` fails, or prematurely cleaned up by higher-level orchestrators (e.g. HyperShift) that do not know the Secret is still referenced by a running VM.

On the first reconcile the Secret is created before the VM, so the ownerReference cannot be set yet. It is added on a subsequent reconcile once the VM is present.

- The existing `reconcileDelete` path continues to explicitly delete the Secret before the VM (deterministic ordering). The ownerReference is a safety net, not a replacement for explicit deletion.
- `pkg/capiv1beta1/map_func.go` has a pre-existing `goimports` formatting change included per CONTRIBUTING.md ("run `make sanity`... commit these files before pushing").
- `make sanity` passes cleanly (linter 0 issues, goimports, 129 unit tests).

```release-note
Set ownerReference on bootstrap userdata Secret linking it to the VirtualMachine for proper lifecycle management
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/controllers/kubevirtmachine_controller.go
+++ b/controllers/kubevirtmachine_controller.go
@@ -30,6 +30,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	kubevirtv1 "kubevirt.io/api/core/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta1" //nolint SA1019
 	capierrors "sigs.k8s.io/cluster-api/errors"          //nolint SA1019
 	"sigs.k8s.io/cluster-api/util"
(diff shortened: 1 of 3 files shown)


amasolov:
Closing this after further analysis. The `-userdata` Secret is shared across all VMs in a NodePool generation, so setting an ownerReference to a single VM would cause premature garbage collection when that particular VM is deleted during a rolling update (while other VMs still reference the same Secret).

The correct fix is upstream in HyperShift: defer the Secret cleanup until the MachineDeployment rollout is complete, rather than adding per-VM ownership at the CAPK layer.

PR #387, opened 2026-07-09
bump all gh actions to their latest versions

nunnatsa:
kubernetes-sigs new policy forbids using floating github-action tags in github workflows, and forces a specific git sha instead.

This PR bump all the imported actions to their latest versions, and uses a git shas instead of floating tags.

> **Note**: the e2e workflow is running the version of the target branch, instead of the version in pull request. The meaning for it is that this action is not fixable using a pull request, because the policy fails this workflow.
>
> To solve it I had to directly push a fix to the main branch; see commit f4d655feb12a3cedc6dbf1e4da3932a984f34aba, where I did the same (replacing the action floating tags with the latest git shas), in the e2e workflow.

**Release notes**
```release-note
None
```

### DIFF
--- a/.github/workflows/clusterkubevirtadm-test.yaml
+++ b/.github/workflows/clusterkubevirtadm-test.yaml
@@ -14,9 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout code
-      uses: actions/checkout@v6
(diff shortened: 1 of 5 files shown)


nunnatsa:
closing and reopening to trigger the e2e tests

PR #392, opened 2026-08-04
OCPBUGS-98217: Attempt live migration before drain+delete for migratable VMIs

orenc1:
When a migratable VMI with `EvictionStrategy=External` is marked for eviction, CAPK now creates a `VirtualMachineInstanceMigration` to attempt zero-disruption live migration before falling back to the disruptive drain+delete path.

The migration lifecycle is handled by a new `tryMigrateVMI()` function:
- Detects successful migration via `NodeName != EvacuationNodeName`
- Waits for in-progress migrations within the existing 10-minute grace period
- Falls back to drain+delete on migration failure or timeout, resetting the grace period so the fallback gets a fresh timeout window
- Falls back gracefully if migration creation fails

Non-migratable VMIs continue through the existing drain+delete path unchanged.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # https://redhat.atlassian.net/browse/OCPBUGS-98217

**Release notes**:

```release-note
Attempt live migration before drain+delete for migratable VMIs
```

### DIFF
--- a/pkg/kubevirt/machine.go
+++ b/pkg/kubevirt/machine.go
@@ -431,10 +431,8 @@ func (m *Machine) Delete() error {
 
 func (m *Machine) DrainNodeIfNeeded(wrkldClstr workloadcluster.WorkloadCluster) (time.Duration, error) {
 	if m.vmiInstance == nil || !m.shouldGracefulDeleteVMI() {
-		if _, anntExists := m.machineContext.KubevirtMachine.Annotations[infrav1.VmiDeletionGraceTime]; anntExists {
-			if err := m.removeGracePeriodAnnotation(); err != nil {
-				return 100 * time.Millisecond, err
-			}
+		if err := m.cleanupMigrationAnnotations(); err != nil {
+			return 100 * time.Millisecond, err
 		}
 
 		// Only uncordon when the VMI is back and not being deleted. If the VMI
@@ -448,6 +446,30 @@ func (m *Machine) DrainNodeIfNeeded(wrkldClstr workloadcluster.WorkloadCluster) 
 		}
 
 		return 0, nil
+	}
+
+	// For migratable VMIs, attempt live migration before the disruptive
+	// drain+delete path. If migration succeeds, the guest node is untouched.
+	if m.vmiInstance.IsMigratable() {
+		migrated, requeueAfter, err := m.tryMigrateVMI()
+		if err != nil {
+			return 0, err
+		}
+		if migrated {
+			if err := m.cleanupMigrationAnnotations(); err != nil {
+				return 100 * time.Millisecond, err
+			}
+			return 0, nil
+		}
+		if requeueAfter > 0 {
+			return requeueAfter, nil
+		}
+		// Migration failed or timed out -- clean up migration annotations
+		// and reset the grace period so the drain+delete fallback gets a
+		// fresh timeout window.
+		if err := m.cleanupMigrationAnnotations(); err != nil {
+			return 100 * time.Millisecond, err
+		}
 	}
 
 	exceeded, err := m.drainGracePeriodExceeded()
@@ -484,6 +506,8 @@ func (m *Machine) DrainNodeIfNeeded(wrkldClstr workloadcluster.WorkloadCluster) 
 
 const removeGracePeriodAnnotationPatch = `[{"op": "remove", "path": "/metadata/annotations/` + infrav1.VmiDeletionGraceTimeEscape + `"}]`
 
+const cleanupMigrationAnnotationsPatch = `{"metadata":{"annotations":{"` + infrav1.VmiMigrationSubmitted + `": null, "` + infrav1.VmiDeletionGraceTime + `": null}}}`
+
 func (m *Machine) removeGracePeriodAnnotation() error {
 	patch := client.RawPatch(types.JSONPatchType, []byte(removeGracePeriodAnnotationPatch))
 
@@ -491,6 +515,24 @@ func (m *Machine) removeGracePeriodAnnotation() error {
 		return fmt.Errorf("failed to remove the %s annotation to the KubeVirtMachine %s; %w", infrav1.VmiDeletionGraceTime, m.machineContext.KubevirtMachine.Name, err)
 	}
 
+	return nil
+}
+
+// cleanupMigrationAnnotations removes both the migration-submitted marker and
+// the grace period annotation in a single merge patch. Setting a key to null
+// in a merge patch removes it; if the key does not exist, it is a no-op.
+func (m *Machine) cleanupMigrationAnnotations() error {
+	annotations := m.machineContext.KubevirtMachine.Annotations
+	_, hasMigration := annotations[infrav1.VmiMigrationSubmitted]
+	_, hasGrace := annotations[infrav1.VmiDeletionGraceTime]
+	if !hasMigration && !hasGrace {
+		return nil
+	}
+
+	patchRequest := client.RawPatch(types.MergePatchType, []byte(cleanupMigrationAnnotationsPatch))
+	if err := m.client.Patch(m.machineContext, m.machineContext.KubevirtMachine, patchRequest); err != nil {
+		return fmt.Errorf("failed to clean up migration annotations on KubevirtMachine %s: %w", m.machineContext.KubevirtMachine.Name, err)
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


tiraboschi:
An orthogonal is to avoid hardcoding `evictionStrategy=External` for CAPK VMs having hypershift NodePools exposing an evictionStrategy that in case of `evictionStrategy=LiveMigrate` will completely bypass CAPK having KubeVirt handling it without triggering CAPK logic at all.
The advantage is that we avoid replicating KubeVirt eviction manager logic (which is more robust with priority queues...) here, the downside is that in case of migration issue, CAPK will never try the drain+delete fallback nor KubeVirt will do.

See: https://github.com/openshift/hypershift/pull/8974

orenc1:
> An orthogonal is to avoid hardcoding `evictionStrategy=External` for CAPK VMs having hypershift NodePools exposing an evictionStrategy that in case of `evictionStrategy=LiveMigrate` will completely bypass CAPK having KubeVirt handling it without triggering CAPK logic at all. The advantage is that we avoid replicating KubeVirt eviction manager logic (which is more robust with priority queues...) here, the downside is that in case of migration issue, CAPK will never try the drain+delete fallback nor KubeVirt will do.
> 
> See: [openshift/hypershift#8974](https://github.com/openshift/hypershift/pull/8974)

Yes, this PR handles the migration on the CAPK level.
It can be a safety net in case the migration done by hypershift at [your PR](https://github.com/openshift/hypershift/pull/8974) is failing indefinitely.
@tiraboschi , do you think this PR makes sense, as a complimentary to the hypershift fix? or should I close it?

tiraboschi:
I think I prefer this one for the sake of consistency: when VMs are handled by CAPK, in my opinion is CAPK that should take care of them (migrating or draining the guest node or deleting) and this is exactly while we have evictionStrategy=External.
Having to conditionally set evictionStrategy=External only when VMs are not migration in my opinion was counter-intuitive.

pkg/kubevirt/machine.go
```diff
@@ -557,6 +584,76 @@ func (m *Machine) uncordonNodeIfNeeded(wrkldClstr workloadcluster.WorkloadCluste
...
+				return false, 0, nil
+			}
+			m.machineContext.Logger.Info("VMI migration in progress, waiting",
+				"vmi", m.vmiInstance.Name)
+			return false, 10 * time.Second, nil
+		}
+	}
+
+	// No active migration -- check if grace period allows starting one
+	exceeded, err := m.drainGracePeriodExceeded()
```
tiraboschi: Maybe the real concern of this approach is if the default drain grace period is large enough to let the bare metal node migrate all the relevant VMs in time on infra clusters runing with many guest clusters.
We have also a limit on the number of parallel outbound migrations per node and https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4426 reduced it to 1 by default to ensure the maximum bandwith for each migration.

orenc1: I think your concern about hitting the timeout is valid, but we're mitigating this scenario by the fact that:

1. KubeVirt's migration controller still governs actual execution regardless of who creates the VirtualMachineInstanceMigration` CR -- CAPK doesn't bypass the queue, it just submits to it.
2. Each VM has its own independent 10-minute timeout, so sequential processing doesn't cause cascading timeouts.
3. If the grace period does expire, the fallback to drain+delete still works -- which is better than the stalling risk in the hypershift fix approach (your PR).

tiraboschi: > I think your concern about hitting the timeout is valid, but we're mitigating this scenario by the fact that:
> 
> 1. KubeVirt's migration controller still governs actual execution regardless of who creates the VirtualMachineInstanceMigration` CR -- CAPK doesn't bypass the queue, it just submits to it.

The evacuation queue is handled by the EvacuationController in Kubevirt, see: https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/watch/drain/evacuation/evacuation.go
Migrations triggered by the evacuation controller are created with:
```
	mig.Spec.Priority = pointer.P(virtv1.PrioritySystemCritical)
	if value, exists := vmi.GetAnnotations()[virtv1.EvictionSourceAnnotation]; exists && value == "descheduler" {
		mig.Spec.Priority = pointer.P(virtv1.PrioritySystemMaintenance)
	}
``` 
while we are not handling migration priority here (off course we can also replicate that logic here but...).

And in addition to the migration queue with priority, the evacuation controller has its own queue, see:
```
		Queue: workqueue.NewTypedRateLimitingQueueWithConfig[string](
			workqueue.DefaultTypedControllerRateLimiter[string](),
			workqueue.TypedRateLimitingQueueConfig[string]{Name: "virt-controller-evacuation"},
		),
```

> 2. Each VM has its own independent 10-minute timeout, so sequential processing doesn't cause cascading timeouts.

This is exactly what scares me:
in a case of a drain of the physical node, the drain controller initiates eviction requests for all eligible pods on the node (including virt-launcher ones) in rapid succession (almost simultaneously).
On KubeVirt this is handled by an admission webhook (synchronous) that, according to the evacuation strategy configured on the VM (see: https://github.com/kubevirt/kubevirt/blob/main/docs/handling-eviction-requests.md#pod-eviction-webhook ) is going to mark VMI for evacuation (or not) accepting the eviction request (or not).
I think that 10 minutes timeout starts there.
On the other side, VM evacuation and migrations are handled asynchronously and sequentially by evacuation and migration controller.
So I fear that with many VMs on the node, the first will be migrated within the 10 minutes boundary while others are simply waiting for the turn while the 10 minutes timeout will expire.

> 3. If the grace period does expire, the fallback to drain+delete still works -- which is better than the stalling risk in the hypershift fix approach (your PR).

in the production code we are now (after my https://github.com/openshift/hypershift/pull/6380 ) hard-coding evictionStrategy=External.
External means that KubeVirt Evacuation controller is not involved at all and capk is only performing guest drain + delete but respecting the drain timeout.

On the other side, if we set evictionStrategy=LiveMigrate, CAPK will do nothing as in `shouldGracefulDeleteVMI` have:
```
	if m.vmiInstance.Spec.EvictionStrategy == nil || *m.vmiInstance.Spec.EvictionStrategy != kubevirtv1.EvictionStrategyExternal {
		m.machineContext.Logger.V(4).Info("DrainNode: graceful deletion is not supported for virtualMachineInstance. Nothing to do here")
		return false
	}
```
but the downside of this is that KubeVirt eviction controller has no attempt nor timeout boundary so it could simply try forever for untranslatable VMs blocking the drain (which is exactly what I naively tried to fix with #6380).

orenc1: regarding the evacuation prioritization, i updated the PR and added `Priority: SystemCritical` to the `tryMigrateVMI()` function to match kubevirt's evacuation controller. This ensures CAPK-created migrations are treated with the same priority in kubevirt's migration queue.

few more things to consider though:

1. each KubevirtMachine has its own independent grace period annotation. the 10-minute timer is per-VM, not shared across the node. so VMs don't compete for the same timeout window.
2. the grace period only starts when CAPK first reconciles the eviction for that specific VM (when setVmiDeletionGraceTime() is called), not when the admission webhook accepts the eviction request. so later VMs in the queue get their own fresh 10-minute window starting from when they're first processed.
3. even in the worst case where migration times out, the VM falls back to drain+delete, which is the existing behavior today. so there's no regression; migration is strictly a best-effort optimization on top.

tiraboschi: > few more things to consider though:
> 
> 1. each KubevirtMachine has its own independent grace period annotation. the 10-minute timer is per-VM, not shared across the node. so VMs don't compete for the same timeout window.
> 2. the grace period only starts when CAPK first reconciles the eviction for that specific VM (when setVmiDeletionGraceTime() is called), not when the admission webhook accepts the eviction request. so later VMs in the queue get their own fresh 10-minute window starting from when they're first processed.

My concern is exactly on this:
Is the pod-eviction-admitter that is (synchronously) marking all the VMs on an node for eviction almost at the same time when the drain controller tries to drain virt-launcher-pod.
https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-api/webhooks/validating-webhook/admitters/pod-eviction-admitter.go#L177

The webhook is going to touch `/status/evacuationNodeName` on each VMI object which, I think, is what triggers a reconciliation here.
Reconciliation are likely processed VM by VM, but really in a short loop so probably the timeouts start, one after the other, for all the VMs on the node in a really short loop (milliseconds or secodns).

So basically we have only 10 minutes to drain, sequentially, all the VMs on a node independelty from the number of VM on the node.

> 3. even in the worst case where migration times out, the VM falls back to drain+delete, which is the existing behavior today. so there's no regression; migration is strictly a best-effort optimization on top.

Fair

orenc1: ok, so to make it more robust, i've updated the PR.
now, we're starting the evacuation timer only when the migration is really started, not when we initiated the evacuation request, by monitoring the `MigrationState` that appears on the VMI.
now all VMs will get a full 10 minutes migration window.
what do you think?

pkg/kubevirt/machine.go
```diff
@@ -560,6 +599,107 @@ func (m *Machine) uncordonNodeIfNeeded(wrkldClstr workloadcluster.WorkloadCluste
...
+	// on a previous reconcile (waiting for KubeVirt to pick it up).
+	if _, exists := m.machineContext.KubevirtMachine.Annotations[infrav1.VmiMigrationSubmitted]; exists {
+		m.machineContext.Logger.Info("Migration already submitted, waiting for KubeVirt to start it",
+			"vmi", m.vmiInstance.Name)
+		return false, 10 * time.Second, nil
+	}
+
+	// First reconcile for this evacuation -- create a migration CR.
+	priority := kubevirtv1.PrioritySystemCritical
+	migration := &kubevirtv1.VirtualMachineInstanceMigration{
```
nunnatsa, line 660: It seems that the required RBAC rule is missing.

pkg/kubevirt/machine.go
```diff
@@ -494,6 +516,23 @@ func (m *Machine) removeGracePeriodAnnotation() error {
 	return nil
 }
 
+// cleanupMigrationAnnotations removes both the migration-submitted marker and
+// the grace period annotation (if they exist). Called when migration succeeds
+// or when falling back to drain+delete.
+func (m *Machine) cleanupMigrationAnnotations() error {
```
nunnatsa, line 524: can we somehow combine the two requests? not sure it's so great to modify the same resource twice in a row.

nunnatsa (commented):
Thanks @orenc1 

Looks good. Please see 2 comments below. The rbac one seems critical to me.

pkg/kubevirt/machine.go
orenc1: good catch.
i had the RBACs but they got removed somehow due to a rebase+force push. fixing.

pkg/kubevirt/machine.go
orenc1: done.
note that i need to add some code to guard the remove annotation patch in case it doesn't exist, to avoid failing the second patch in the same request.

nunnatsa: We can use merge patch instead, and use it as many times as we want with no error. Not even need the escape in this case.

This should work:
```
`{"metadata":{"annotations":{"` + infrav1.VmiMigrationSubmitted + `": null, "` + infrav1.VmiDeletionGraceTime + `": null}}}`
```

pkg/kubevirt/machine.go
```diff
@@ -494,6 +516,18 @@ func (m *Machine) removeGracePeriodAnnotation() error {
 	return nil
 }
 
+// cleanupMigrationAnnotations removes both the migration-submitted marker and
+// the grace period annotation in a single merge patch. Setting a key to null
+// in a merge patch removes it; if the key does not exist, it is a no-op.
+func (m *Machine) cleanupMigrationAnnotations() error {
+	patch := `{"metadata":{"annotations":{"` + infrav1.VmiMigrationSubmitted + `": null, "` + infrav1.VmiDeletionGraceTime + `": null}}}`
```
nunnatsa: please make it `const`

orenc1: done.

pkg/kubevirt/machine.go
orenc1: done with merge patch instead of the cumbersome json patch.

nunnatsa (approved):
/approve

PR #393, opened 2026-08-04
Add cloud-init data source and network data configuration to KubevirtMachine

chdxD1:

CAPK builds the cloud-init config drive for every virtual machine and sets only `UserDataSecretRef` on it. `NetworkDataSecretRef` is never set, so there is no way to hand a static network configuration to the guest and guests depend on DHCP.

The `virtualMachineTemplate` passthrough does not help here. `buildVirtualMachineInstanceTemplate` deep copies the user supplied template and then appends its own cloud-init volume, so a cloud-init volume supplied by the user collides with the one CAPK adds. Pre-creating the bootstrap secret in the infra cluster does not help either, because the controller reconciles it with `CreateOrUpdate` and replaces `Data` wholesale.

**Implementation**

This adds an optional `cloudInit` section to `KubevirtMachineSpec`:

```yaml
kind: KubevirtMachine
spec:
  cloudInit:
    dataSource: noCloud
    networkDataSecretRef:
      name: my-machine-network-data
```

When `networkDataSecretRef` is set, the referenced secret is read from the KubevirtMachine's namespace and its network data is copied into the `<bootstrap-secret>-userdata` secret that CAPK already creates in the infra cluster. The cloud-init volume then points at that secret for both user data and network data.

**Why there is a data source selector**

cloud-init parses the network configuration document differently depending on the data source:

| Data source | File | Accepted format |
| --- | --- | --- |
| `configDrive` | `openstack/latest/network_data.json` | OpenStack `network_data.json` only, or ENI |
| `noCloud` | `network-config` | cloud-init network config v1 or v2 |

`DataSourceConfigDrive.network_config` only ever calls `openstack.convert_net_json` or `eni.convert_eni_data`, so a config drive can never accept netplan style YAML.

The selector defaults to `configDrive`, so existing machines keep their current data source and behaviour is unchanged when `cloudInit` is not set.

Note that user data is unaffected by the selection. The file name differs between data sources, but the payload is the same cloud-config or script in both cases.

The one thing `noCloud` gives up is the config drive `devices` metadata, which carries Nova style MAC and tag information. `NoCloudMetadata` has no equivalent.

**Relationship to #386**

#386 by @qinqon approaches the same gap from the HyperShift side, using an annotation carrying inline network data and switching the volume to NoCloud. That PR notes that inline network data is capped at 2048 bytes by the KubeVirt admitter, that a `networkDataSecretRef` variant is future work, and that an upstreamable version needs an opt in gate such as a typed API field.

This PR is intended to be that typed API field. It has no size cap because the document lives in a secret, and it makes the data source explicit rather than switching it unconditionally. Happy to align the two so the annotation path can build on this.

**Scope**

The provider stays a plain carrier for the document. Rendering the network configuration and allocating addresses stay with whoever creates the KubevirtMachine. Future work might be to include the upstream CAPI IPAM contract.

**Open questions for reviewers**

1. `LocalObjectReference` restricts the secret to the KubevirtMachine's namespace. `cluster-api-provider-metal3` uses a `SecretReference` and allows cross namespace. Happy to switch if cross namespace is wanted.
2. The network data is stored in the existing `<bootstrap-secret>-userdata` secret, which now carries more than user data. A separate secret would be cleaner in name but would add another object to reconcile and garbage collect.

**Which issue(s) this PR fixes**:

None.

Still a draft, opened to get direction on the API shape before filling in docs.

**Release note**:

```release-note
KubevirtMachine gained an optional `cloudInit` section with a `dataSource` selector (`configDrive` or `noCloud`, defaulting to `configDrive`) and a `networkDataSecretRef`, allowing a static cloud-init network configuration to be attached to the virtual machine instead of relying on DHCP.
```


chdxD1:
now sure why I was so blind - there is https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/pull/386 as well

It also has the same issues in terms of NoCloud / ConfigDrive compatibility.

qinqon:
@chdxD1 Hi, going to take a look on this, my PoC was just to test something, this is better a approach, let's see if it fits on what we need.

api/v1alpha1/kubevirtmachine_types.go
```diff
@@ -55,6 +55,69 @@ type KubevirtMachineSpec struct {
 	// When nil, this defaults to the value present in the KubevirtCluster object's spec associated with this machine.
 	// +optional
 	InfraClusterSecretRef *corev1.ObjectReference `json:"infraClusterSecretRef,omitempty"`
+
+	// CloudInit configures the cloud-init data source attached to the virtual machine.
+	// When nil, a config drive data source carrying only the bootstrap user data is attached.
+	// +optional
+	CloudInit *CloudInitSpec `json:"cloudInit,omitempty"`
```
qinqon, line 62: Lifecycle semantics: the `-userdata` secret is reconciled with `CreateOrUpdate` on every pass, so edits to the referenced secret (or removing the ref) propagate to the infra secret, but the VM is never updated after `Create`. Removing `networkDataSecretRef` after the VM exists drops the `networkdata` key while the VMI volume still references it, so a VMI restart fails; changing content has no effect until next boot.

I'd document that `cloudInit` applies at first boot only and treat it as effectively immutable, or add a CEL `self == oldSelf` on it like `KubevirtClusterTemplate` does for its spec.

api/v1alpha1/kubevirtmachine_types.go
```diff
@@ -55,6 +55,69 @@ type KubevirtMachineSpec struct {
...
+	// DataSource selects the type of cloud-init data source attached to the virtual machine.
+	// This choice determines the format expected for the network configuration document, because
+	// cloud-init parses it differently per data source: "configDrive" expects the OpenStack
+	// network_data.json format, while "noCloud" expects the cloud-init network configuration
+	// format, version 1 or 2.
+	// Defaults to "configDrive".
+	// +optional
+	// +kubebuilder:validation:Enum=configDrive;noCloud
+	// +kubebuilder:default:=configDrive
+	DataSource CloudInitDataSource `json:"dataSource,omitempty"`
```
qinqon, line 89: nit: worth noting in the godoc that `noCloud` also changes where user data lives (`/user-data` vs `/openstack/latest/user_data`), so operators know their Ignition/cloud-init reader must support the NoCloud layout too.

api/v1alpha1/kubevirtmachine_types.go
```diff
@@ -55,6 +55,69 @@ type KubevirtMachineSpec struct {
...
+	// NetworkDataSecretRef is a reference to a secret in the KubevirtMachine's namespace, holding a
+	// network configuration document under a "networkdata" or "networkData" key, in the format
+	// expected by DataSource.
+	// When set, the content is copied to the infra cluster and attached to the cloud-init data
+	// source of the virtual machine, allowing the guest to configure its network without relying
+	// on DHCP. This is intended for setups where addresses are managed externally, for example on
+	// secondary interfaces attached through Multus.
+	// When nil, the guest is expected to obtain its network configuration by other means, such as DHCP.
+	// +optional
+	NetworkDataSecretRef *corev1.LocalObjectReference `json:"networkDataSecretRef,omitempty"`
```
qinqon, line 100: Two things here:

1. Please add an inline `networkData` alternative, mutually exclusive with the ref, mirroring KubeVirt's own `CloudInitNoCloudSource`/`CloudInitConfigDriveSource` which offer both `networkData` and `networkDataSecretRef`:

```yaml
cloudInit:
  dataSource: noCloud
  networkData: |
    version: 2
    ...
```

with `+kubebuilder:validation:XValidation:rule="!(has(self.networkData) && has(self.networkDataSecretRef))"`. The document isn't secret material (gateways, DNS, MAC/PCI match). Since the controller would still copy the inline bytes into the `-userdata` secret rather than KubeVirt's inline field, the 2048-byte admitter cap from #386 doesn't apply. For a producer like HyperShift that renders this per NodePool from a typed API, inline avoids an extra Secret to create/hash/GC and keeps the `KubevirtMachineTemplate` self-contained so template name hashing gives correct rollout semantics for free. Keeping the ref for large/IPAM-generated documents still makes sense.

2. `LocalObjectReference.Name` defaults to `""` in the generated CRD, so `networkDataSecretRef: {}` passes admission and fails later in `r.Get` with an opaque error. A `+kubebuilder:validation:MinLength=1` on the name (or a CEL rule) would catch it at admission.

controllers/kubevirtmachine_controller.go
```diff
@@ -671,6 +676,9 @@ func (r *KubevirtMachineReconciler) reconcileKubevirtBootstrapSecret(ctx *contex
 		newBootstrapDataSecret.Data = map[string][]byte{
 			"userdata": value,
 		}
+		if len(networkData) > 0 {
```
qinqon, line 679: Bug: `pkg/kubevirt/utils.go` sets `NetworkDataSecretRef` whenever `GetNetworkDataSecretRef() != nil`, regardless of content. If the referenced secret has the key with an empty value, this branch skips writing `networkdata` while the VMI volume still references it, and virt-launcher fails to start the VMI.

Either always write the key when a ref is set, or (better) return an error from `getNetworkData` on empty content.

nit: pre-existing, but the `OperationResultCreated/Updated` log lines below say "capk user with ssh config" and the secret now also carries network data; maybe generalize the message.

controllers/kubevirtmachine_controller.go
```diff
@@ -689,6 +697,30 @@ func (r *KubevirtMachineReconciler) reconcileKubevirtBootstrapSecret(ctx *contex
...
+	}
+
+	// Both spellings are accepted, matching the keys KubeVirt itself looks up in a network data secret.
+	for _, dataKey := range []string{"networkdata", "networkData"} {
+		if networkData, ok := s.Data[dataKey]; ok {
+			return networkData, nil
+		}
+	}
+
+	return nil, errors.Errorf("error retrieving network data: secret %s is missing the networkdata key", key)
```
qinqon, line 721: A missing/invalid network-data secret surfaces to the caller as `WaitingForBootstrapDataReason`. A dedicated reason (e.g. `WaitingForNetworkDataReason`) would make it easier to tell "no CAPI bootstrap secret yet" from "user pointed at a bad secret".

Also, `SetupWithManager` doesn't watch `corev1.Secret`, so a secret created shortly after the `KubevirtMachine` is only picked up via the 10s `RequeueAfter`. Acceptable, but a `Watches(&corev1.Secret{}, ...)` mapping back to referencing machines would make the create-order race quieter (less relevant if inline is added).

pkg/kubevirt/utils.go
```diff
@@ -142,15 +143,32 @@ func buildVirtualMachineInstanceTemplate(ctx *context.MachineContext) *kubevirtv
 	template.Spec = *ctx.KubevirtMachine.Spec.VirtualMachineTemplate.Spec.Template.Spec.DeepCopy()
 
 	cloudInitVolumeName := "cloudinitvolume"
+	cloudInitSecretName := *ctx.Machine.Spec.Bootstrap.DataSecretName + "-userdata"
+	userDataSecretRef := &corev1.LocalObjectReference{Name: cloudInitSecretName}
+
+	// The network data is copied into the same secret as the user data. KubeVirt mounts each
+	// reference into its own pod volume, so pointing both references at one secret is safe.
```
qinqon, line 150: This claim is what an e2e should cover: that KubeVirt accepts both `userDataSecretRef` and `networkDataSecretRef` pointing at the same secret and virt-launcher starts the VMI. The unit tests only assert the generated `VirtualMachine` shape.

controllers/kubevirtmachine_controller_test.go
```diff
@@ -388,6 +388,153 @@ var _ = Describe("reconcile a kubevirt machine", func() {
...
+
+		noCloud := cloudInitNoCloud(vm)
+		Expect(noCloud).ShouldNot(BeNil())
+		Expect(noCloud.UserDataSecretRef).ShouldNot(BeNil())
+		Expect(noCloud.UserDataSecretRef.Name).To(Equal(bootstrapDataSecretKey.Name))
+		Expect(noCloud.NetworkDataSecretRef).ShouldNot(BeNil())
+		Expect(noCloud.NetworkDataSecretRef.Name).To(Equal(bootstrapDataSecretKey.Name))
+	})
+
+	It("should fail when the referenced network data secret has no networkdata key", func() {
```
qinqon, line 507: Please also add the empty-value case (`networkdata: ""`) and, once validation is in place, the ref-with-empty-name case.

qinqon (changes requested):

PR #395, opened 2026-08-11
Fix nil map panics and set CAPI v1beta2 initialization status

orenc1:

Fix nil map assignment panics when creating VMs from templates without annotations. The `Create() mutateFn` and `newVirtualMachineFromKubevirtMachine` both write to the annotations map without checking if it's nil.

Also set `status.initialization.provisioned=true` in both the cluster and machine controllers when the resource becomes Ready. This field is required by the CAPI v1beta2 Machine controller to proceed with `providerID` propagation.

Additionally, use the typed `kubevirtv1.DiskBusVirtio` constant instead of the hardcoded `"virtio"` string for the cloud-init disk bus.

**Release notes**:

```release-note
Fix nil map panics and set CAPI v1beta2 initialization status
```


nunnatsa (approved):
/approve

PR #397, opened 2026-08-19
Bump kubevirt.io/api to v1.9.0 to expose network DRA fields

### LINKED ISSUE #396: kubevirt.io/api pinned to v1.8.0 keeps networks[].resourceClaim (KubeVirt 1.9 network DRA) out of the generated CRDs
**User Story**

As an operator I would like to request DRA-backed SR-IOV NICs for the machines CAPK provisions, so that I can run RDMA/GPU workloads on CAPK-managed clusters the same way I already can on VMs I create directly with KubeVirt.

**Detailed Description**

KubeVirt 1.9 added alpha support for DRA-backed networks through the VMI field `spec.networks[].resourceClaim`, behind the `NetworkDevicesWithDRA` feature gate (kubevirt/kubevirt#17661). The same release promoted `GPUsWithDRA` and `HostDevicesWithDRA` to beta (kubevirt/kubevirt#18446).

`go.mod` on `main` still pins:

```
kubevirt.io/api v1.8.0
```

Because `VirtualMachineTemplateSpec` embeds `kubevirtv1.VirtualMachineSpec` wholesale rather than restating fields:

https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/main/api/v1alpha1/kubevirtmachine_types.go#L34-L40

the pin propagates straight into the generated CRDs. On current `main`, the `networks[]` items of `kubevirtmachines` and `kubevirtmachinetemplates` only offer:

```
['multus', 'name', 'pod']
```

So even on a cluster where `NetworkDevicesWithDRA` is enabled and DRA-backed NICs work for directly-created VMs, there is no way to request one for a machine created by CAPK — the field is rejected/pruned before it ever reaches KubeVirt.

**What the bump actually changes**

I bumped `kubevirt.io/api` to v1.9.0 locally and re-ran `make generate`. No Go code change was needed. Results:

- `go build ./...`, `go vet ./...`, `make test`, `make manager` all pass unchanged — v1.8.0 → v1.9.0 introduces no breaking change in the surface CAPK consumes.
- `make check-gen` is clean, and the only regenerated artifacts are the two CRDs.
- Diffing every schema path before/after, the change is **purely additive — zero properties removed**:

```
+ .../networks/resourceClaim
+ .../networks/resourceClaim/claimName
+ .../networks/resourceClaim/requestName
+ .../domain/devices/interfaces/portRanges{start,end,protocol}
+ .../template/spec/serviceAccountName
```

**Backwards compatibility**

`Network.ResourceClaim` is `*ClaimRequest` with `omitempty`, as are the other new fields. A manifest that does not use them serializes byte-for-byte as it did under v1.8.0, so CAPK built against v1.9.0 keeps working against older KubeVirt clusters; only users who opt into the new fields need a 1.9 infra cluster with the gate enabled.

Worth noting that the e2e lane already installs KubeVirt from `kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/latest` (`hack/run-e2e.sh`), i.e. CI is already exercised against a KubeVirt newer than the pinned API module.

**Verification**

Beyond the build/test/lint/check-gen runs, I stood up an envtest API server with the regenerated CRDs and applied a `KubevirtMachineTemplate` carrying both `spec.resourceClaims[]` and `networks[].resourceClaim`. Both survived the round-trip without pruning:

```
OK: networks[0].resourceClaim preserved = map["claimName":"gpu-nic" "requestName":"nic-request"]
OK: spec.resourceClaims preserved = [map["name":"gpu-nic" "resourceClaimTemplateName":"sriov-nic-template"]]
```

Against the current pinned v1.8.0 CRDs the same manifest loses `networks[0].resourceClaim` to pruning.

I appreciate that network DRA is still alpha in KubeVirt, so if the preference is to wait for it to reach beta/GA before exposing it, that is completely understandable. Two things that may argue for bumping now regardless: the bump also picks up the GPU/HostDevice DRA promotion to beta, and since the spec is embedded, keeping the module pinned holds back every new KubeVirt API field, not just the DRA ones.

Happy to open a PR — I have the change ready and passing locally.

/kind feature

changjjjjjjj:

Bumps `kubevirt.io/api` from v1.8.0 to v1.9.0 and regenerates the manifests, so that the KubeVirt 1.9 API fields — most notably `spec.networks[].resourceClaim` for DRA-backed networks — become reachable through `KubevirtMachine` / `KubevirtMachineTemplate`.

Because `VirtualMachineTemplateSpec` embeds `kubevirtv1.VirtualMachineSpec` wholesale rather than restating fields, the module pin propagates directly into the generated CRDs. With v1.8.0 the `networks[]` items only offered `multus` / `name` / `pod`, so a DRA-backed NIC could not be requested for any machine CAPK creates even when `NetworkDevicesWithDRA` was enabled on the infra cluster. The bump also picks up the promotion of `GPUsWithDRA` / `HostDevicesWithDRA` to beta.

No Go code change is required — only `go.mod`/`go.sum` and the two regenerated CRDs.

**Which issue this PR fixes**: fixes #396

*The change is additive.* I diffed every schema path in both CRDs before and after regeneration. **No property is removed**; the additions are:

```
+ .../networks/resourceClaim{claimName,requestName}          <- network DRA
+ .../domain/devices/interfaces/portRanges{start,end,protocol}
+ .../template/spec/serviceAccountName
```

*Backwards compatibility.* `Network.ResourceClaim` is `*ClaimRequest` with `omitempty`, as are the other new fields, so a manifest that does not use them serializes exactly as it did under v1.8.0. CAPK built against v1.9.0 continues to work against older KubeVirt clusters; only users opting into the new fields need a 1.9 infra cluster with the gate enabled.

*Checks run locally* (macOS/arm64, go 1.26.2):

| check | result |
|---|---|
| `go build ./...` / `go vet ./...` | pass |
| `make test` | pass |
| `make manager` | pass |
| `make clusterkubevirtadm-test` | pass |
| `golangci-lint run --timeout=5m` (v2.11.2, the version CI uses) | `0 issues` |
| `make check-gen` | clean (exit 0) |
| `make sanity` | exit 0 |

*Extra verification.* I stood up an envtest API server with the regenerated CRDs and applied a `KubevirtMachineTemplate` carrying both `spec.resourceClaims[]` and `networks[].resourceClaim`. Both survive the round-trip without pruning:

```
OK: networks[0].resourceClaim preserved = map["claimName":"gpu-nic" "requestName":"nic-request"]
OK: spec.resourceClaims preserved = [map["name":"gpu-nic" "resourceClaimTemplateName":"sriov-nic-template"]]
```

The same manifest loses `networks[0].resourceClaim` to pruning against the v1.8.0-generated CRDs. I kept that scratch test out of the PR since the repo has no envtest harness today and `make test` would need `KUBEBUILDER_ASSETS` to run it — happy to contribute it properly if you would like the coverage.

*Two things I did not touch, but noticed:*
- `make sanity` runs `goimports`, which rewrites four `api/v1alpha1/*_types.go` files (trailing blank line). This reproduces on a clean `upstream/main` checkout with no bump applied, so it is pre-existing and I left it out of this PR rather than mixing in unrelated diff.
- `hack/kccm-flavor-gen.sh` uses GNU `sed -i -E` semantics and fails on macOS BSD sed, so `make generate` cannot complete on a Mac without `gsed`. Also pre-existing and out of scope here; glad to file it separately.

*On the alpha status:* network DRA is alpha in KubeVirt 1.9, so if you would rather wait for beta before exposing it, that is understandable. Since the spec is embedded, though, holding the pin holds back every new KubeVirt API field rather than just the DRA ones — and `hack/run-e2e.sh` already installs KubeVirt from the nightly `latest` channel, so CI runs against a KubeVirt newer than the pinned module either way.

**Release notes**:
```release-note
Bump kubevirt.io/api to v1.9.0, exposing the KubeVirt 1.9 API fields in KubevirtMachine and KubevirtMachineTemplate — including `spec.networks[].resourceClaim` for DRA-backed networks (alpha, requires the NetworkDevicesWithDRA feature gate on the infra cluster).
```


nunnatsa (approved):
/ok-to-test

PR #399, opened 2026-08-27
Enforce same-namespace constraint on infraClusterSecretRef and tighten RBAC

orenc1:
Ensure that `infraClusterSecretRef` always resolves within the owning resource's namespace. Previously, a non-empty `.namespace` field was honored verbatim, which could lead to unintended behavior when the field referenced a different namespace.

Changes:
- Add runtime validation in `GenerateInfraClusterClient` that rejects references pointing outside the owner namespace.
- Register validating webhooks for `KubevirtCluster` and `KubevirtMachine` that reject cross-namespace `infraClusterSecretRef` on create/update.
- Reduce RBAC verbs on secrets (drop `delete`, `patch`) and services (drop patch) to match actual controller usage.
- Regenerate CRD and webhook manifests.

**Release notes**:

```release-note
Enforce same-namespace constraint on infraClusterSecretRef and tighten RBAC
```

### DIFF
--- a/pkg/webhookhandler/infrasecretref_validator.go
+++ b/pkg/webhookhandler/infrasecretref_validator.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhookhandler
+
+import (
+	"context"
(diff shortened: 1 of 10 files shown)


pkg/webhookhandler/infrasecretref_validator.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return admission.Allowed("")
+}
+
+func validateInfraClusterSecretRef(ref *corev1.ObjectReference, resourceNamespace string) error {
+	if ref == nil {
+		return nil
+	}
+	if ref.Namespace != "" && ref.Namespace != resourceNamespace {
```
nunnatsa: Q: What is the use-case of an empty namespace?

orenc1: when `infraClusterSecretRef.Namespace` is empty (`""`), it falls back to the namespace of the owning resource (either KubevirtCluster or KubevirtMachine).

This is the common pattern in Kubernetes APIs -- for example, `corev1.ObjectReference` documents `.Namespace` as optional, and when omitted, consumers assume the same namespace as the referencing object

nunnatsa (approved):
/approve
/lgtm

PR #401, opened 2026-09-01
Add KubevirtRemediation for restart-based external remediation

### LINKED ISSUE #398: Support CAPI external remediation so unhealthy Machines can be restarted instead of replaced
**User Story**

As an operator I would like an unhealthy KubeVirt-backed Machine to be restarted rather than deleted and rebuilt, so that faults which only need a reboot do not cost a full node rebuild.

**Detailed Description**

CAPI already defines the extension point for this: when `MachineHealthCheck.spec.remediation.templateRef` is set, the MHC creates a provider-owned remediation object instead of deleting the Machine, and the provider reports back through the `MachineExternallyRemediated` condition. CAPI core deliberately has no reboot primitive; it delegates to providers.

Metal3 implements this today (`api/v1beta1/metal3remediation_types.go`, `api/v1beta2/metal3remediation_types.go`), with a reboot strategy backed by BMC power cycling.

CAPK has no remediation types at all. The only remediation-adjacent code is `KubevirtMachineVMTerminalLabel` in `pkg/kubevirt/machine.go`, which marks a VM terminal so that CAPI's normal failure path deletes and replaces it — the opposite of a restart. So for KubeVirt-backed nodes, the only remediation CAPI can perform is a full Machine replacement.

The gap becomes visible as soon as an external health system is involved. We run NVIDIA NVSentinel inside workload clusters whose nodes are KubeVirt VMs with GPUs attached by passthrough. Its most severe GPU fault — Xid 79, "GPU has fallen off the bus" — comes with a recommended action of *restart the machine*, not replace it. There is nothing in CAPK for that recommendation to map onto, so the available choices are rebuilding the node (image pull, kubeadm join, addon rollout, and an idle passthrough GPU for the duration) or doing nothing at all.

Proposed shape, following Metal3:

- `KubeVirtRemediation` / `KubeVirtRemediationTemplate` with a `Reboot` strategy, `retryLimit`, and `timeoutSeconds`.
- Back it with KubeVirt's existing restart primitive (the `virtualmachines/restart` subresource, equivalent to `virtctl restart`).
- Track progress in `.status.phase`, and on retry/timeout exhaustion set `MachineOwnerRemediated=False` so CAPI falls back to replacing the Machine.

Two things make this a natural fit:

- A KubeVirt VM restart **preserves the node name**, which matches what CAPI expects from a remediation that is not a replacement.
- The CRD shape and controller contract already exist in Metal3, so there is little new design surface to argue about.

Observed with CAPI v1.13.3 and CAPK on a KubeVirt-backed workload cluster.

I would like to implement this, and would appreciate a maintainer's view on the proposed shape before I open a PR.

/kind feature

changjjjjjjj:

Implements Cluster API's [external remediation](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/healthchecking) contract for CAPK, so a `MachineHealthCheck` can restart the VirtualMachineInstance backing an unhealthy Machine instead of deleting and rebuilding the whole node.

New API types `KubevirtRemediation` / `KubevirtRemediationTemplate` (shape follows `Metal3Remediation` in cluster-api-provider-metal3: a `Reboot` strategy with `retryLimit` and `timeoutSeconds`), plus a controller that reconciles the remediation objects the MachineHealthCheck controller instantiates from `spec.remediation.templateRef`:

1. The controller deletes the Machine's VMI. With `runStrategy: Always` (set by all CAPK cluster templates) the KubeVirt VM controller recreates it immediately — same VirtualMachine, same disks, same node name. This is the API-level equivalent of a power cycle. The `virtctl restart` subresource is not reachable through controller-runtime, which is why VMI deletion is used.
2. If the node passes its health check again, the MachineHealthCheck controller deletes the remediation object and reconciliation stops — success needs no extra signalling.
3. If the Machine is still unhealthy after `timeoutSeconds`, the controller restarts again, up to `retryLimit` times.
4. When the retry budget is exhausted — or a restart cannot bring the node back at all — the controller sets the `OwnerRemediated` condition to `False`, and the MachineSet falls back to the usual delete-and-replace.

A restart is only attempted when it can bring the node back:

- The VM's run strategy must be `Always`; otherwise a deleted VMI would stay gone and a restart would become an outage.
- The node's state must live on persistent storage. A VM with any `containerDisk` or `ephemeral` volume is handed back to Cluster API without touching the VMI: the recreated VMI would boot from a pristine image without kubelet credentials (or local etcd data), and the bootstrap token needed to join again has usually expired. Most templates in `templates/` use `containerDisk`; for those, external remediation degrades to plain replacement. `templates/cluster-template-persistent-storage.yaml` is the shape that benefits.
- Transient errors (e.g. an unreachable infrastructure cluster) are retried with backoff and never cause a replacement on their own.

Pausing is honored: while `Cluster.spec.paused` is set, or the Machine or the KubevirtRemediation carries the paused annotation, the controller neither restarts nor escalates, even with a retry pending. Since the MachineHealthCheck controller is paused for the same time, its verdict is stale when the pause is lifted; the controller waits one minute after the `Paused` conditions of the Cluster and the Machine turn `False`, and it never retries or escalates while `HealthCheckSucceeded` is `True`.

The controller is opt-in (`--enable-remediation`, default off), so installations that upgrade the controller image separately from the CRDs are not affected: with the flag unset the manager does not watch the new kinds and runs without the CRDs.

Motivation: for a class of node faults a reboot is the recommended and sufficient fix (our concrete case: GPU passthrough nodes where Xid 79 "GPU has fallen off the bus" is remediated by a restart). Rebuilding the node instead costs image pulls, kubeadm join and addon rollout, while the passthrough GPU sits idle.

**Which issue this PR fixes**: fixes #398

**Testing**:

- Unit tests for the controller (fake client; pause, grace period, health gate, disk guard, transient vs. terminal errors, hand-over).
- `make test-integration` (new, envtest; CI job `envtest`): the manager runs without the remediation CRDs while the flag is off and its controllers reconcile; enabling the flag without the CRDs fails the setup with a no-match error; with the CRDs and the flag on, a request is left alone while its Cluster is paused and picked up on unpause (which exercises the watches). `setupReconcilers` returns its error instead of calling `os.Exit` to make this testable.
- e2e (`Label("remediation")`, the MachineHealthCheck is driven by a node condition owned by the tests, so a spec decides when a node is unhealthy and when it has recovered):
  - restart in place on persistent storage, with the cluster paused while a retry is pending: no retry although the boot window expires; after clearing the fault and unpausing, the request is removed and the node is not restarted again; same Machine, VM, PVCs, Node and machine-id, new VMI UID and boot ID, Node Ready.
  - retry exhaustion followed by Machine replacement, on external infrastructure with persistent storage.
  - a Machine booting from a `containerDisk` (default template) is replaced right away, without a restart.
  - The `config/kubevirtci` overlay enables the controller. The suite now runs for a little over an hour, so `hack/run-e2e.sh` raises Ginkgo's default one hour suite timeout to 2h (the prow job allows 3h).

- Escalation sets the v1beta2 `OwnerRemediated` condition (`conditions.Set` with `metav1.Condition`), which is what the MachineSet's `IsUnhealthyAndOwnerRemediated` filter reads in CAPI >= v1.13. Metal3 still uses the deprecated v1beta1 condition helpers for the same purpose.
- Control plane Machines are remediated by their owner controller (e.g. KubeadmControlPlane), which applies its own safety rules; this controller does not special-case them.
- RBAC added on top of main: `machines/status` (update, patch) for the `OwnerRemediated` condition, and the new `kubevirtremediations` kind plus its status. VM/VMI access reuses the existing rules (get / delete).
- `make generate` / `make check-gen` are clean. `make sanity` rewrites the trailing newline of five pre-existing files on untouched main as well; those are left out of this PR.

**Release notes**:

```release-note
Added KubevirtRemediation and KubevirtRemediationTemplate: MachineHealthCheck external remediation that restarts the VirtualMachineInstance of an unhealthy Machine (with retry limit and timeout) before falling back to Machine replacement. Opt-in via the --enable-remediation manager flag; requires persistent VM storage.
```


changjjjjjjj:
/cc @nunnatsa 
Hi. Can you review this PR plz?

changjjjjjjj:
@qinqon Hi, It's been 3 weeks after I opened this PR. Would you give a quick review of this?

changjjjjjjj:
@qinqon Thanks for your comment! However, I think we need /ok-to-test rather than /test all

main.go
```diff
@@ -252,6 +252,14 @@ func setupReconcilers(ctx context.Context, mgr ctrl.Manager) {
 		os.Exit(1)
 	}
 
+	if err := (&controllers.KubevirtRemediationReconciler{
+		Client:       mgr.GetClient(),
+		InfraCluster: infracluster.New(mgr.GetClient(), noCachedClient, controllerNS),
+	}).SetupWithManager(ctx, mgr); err != nil {
```
qinqon, line 258: For installations that manage CRDs separately from the controller image, an image-only upgrade leaves KubevirtRemediation unavailable. This unconditional watch then times out and stops the entire manager, even when remediation is unused (reproduced locally). Please make controller activation optional and document that both remediation CRDs must be installed before enabling it.

controllers/kubevirtremediation_controller.go
```diff
@@ -0,0 +1,280 @@
...
+// object and reconciliation simply stops.
+type KubevirtRemediationReconciler struct {
+	client.Client
+	InfraCluster infracluster.InfraCluster
+}
+
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=kubevirtremediations,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=kubevirtremediations/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=kubevirtremediationtemplates,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines,verbs=get;list;watch;update;patch
```
qinqon, line 63: handOverToCAPI patches Machine status, but this marker and the generated role omit machines/status. With CAPK's shipped RBAC, escalation fails with Forbidden and replacement is never triggered. Please grant status patch/update permissions and regenerate RBAC.

controllers/kubevirtremediation_controller.go
```diff
@@ -0,0 +1,280 @@
...
+		// Only Always guarantees the VM controller brings the VMI back after
+		// we delete it. Under Manual or Halted a deleted VMI stays gone, which
+		// would silently turn a restart into an outage.
+		return fmt.Errorf("VM run strategy is %q, restart requires %q", runStrategy, kubevirtv1.RunStrategyAlways)
+	}
+
+	vmi := &kubevirtv1.VirtualMachineInstance{}
+	vmi.Namespace = vmKey.Namespace
+	vmi.Name = vmKey.Name
+	if err := infraClient.Delete(ctx, vmi); err != nil && !apierrors.IsNotFound(err) {
```
qinqon, line 215: Deleting a VMI discards containerDisk writable state. The default CAPK template uses containerDisk, so this loses kubelet credentials/local etcd data rather than preserving the node; an expired bootstrap token can also prevent rejoining. Please verify state-bearing disks persist before restarting, fall back to replacement otherwise, and document the persistent-storage requirement.

controllers/kubevirtremediation_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	if machine == nil {
+		// The MachineHealthCheck controller always sets the owner reference at
+		// creation time, so this only happens for hand-made objects.
+		logger.Info("KubevirtRemediation has no owner Machine, ignoring")
+		return ctrl.Result{}, nil
+	}
+	logger = logger.WithValues("machine", machine.Name)
+
+	if !machine.DeletionTimestamp.IsZero() {
```
qinqon, line 99: Please honor Cluster.spec.paused and pause annotations before restarting or escalating. If the cluster is paused after remediation starts, timed retries still delete the VMI. MHC reconciliation is paused too, so the request can remain even after recovery and trigger an unnecessary restart during maintenance. Ensure reconciliation resumes after unpausing.

controllers/kubevirtremediation_controller_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	It("ignores a remediation without an owner Machine", func() {
+		remediation.OwnerReferences = nil
+		setup(remediation, machine, kubevirtMachine, vm, vmi)
+
+		result, err := reconcile()
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(result.RequeueAfter).To(BeZero())
+	})
+})
```
qinqon, line 312: Could you add e2e coverage for successful reboot/recovery and retry exhaustion followed by Machine replacement, including external infrastructure? For recovery, assert the VMI UID changes while the VM/PVC identities and node state are preserved, the Node becomes Ready, and the remediation request is removed. Please also cover pausing during a pending retry. A separate manager-startup integration test should verify existing controllers still run without remediation CRDs when the feature is disabled.

qinqon (changes requested):
Four issues: startup compatibility when CRDs are managed separately, status RBAC, ephemeral disk state loss, and pause handling. Focused tests passed; missing-CRD manager failure reproduced locally. Live recovery remains untested.

main.go
changjjjjjjj: Good catch, thanks. The controller is now opt-in: `--enable-remediation` (default `false`) gates `SetupWithManager`, so with the flag unset the manager never watches the new kinds and runs fine without the CRDs. `docs/external-remediation.md` has a new "Enabling the controller" section that says both CRDs must be installed before turning the flag on, and why.

controllers/kubevirtremediation_controller.go
changjjjjjjj: You're right — the fake client in the unit tests doesn't enforce RBAC, so this slipped through. Added `machines/status` with `update;patch` and regenerated the role. While there I trimmed the markers to what the controller actually uses: `machines` stays read-only (the condition goes through `Status().Patch` only), VM is `get`, VMI is `delete` (the infra client is uncached, so no list/watch), and the unused `kubevirtremediationtemplates` rule is gone. Net RBAC change versus main is now `machines/status` plus the new `kubevirtremediations` kind and its status.

controllers/kubevirtremediation_controller.go
changjjjjjjj: Agreed, and the "worst case is identical to today" claim in the description was wrong for containerDisk — fixed there too. `restartVM` now inspects the VM's volumes before deleting anything: if any volume is a `containerDisk` or `ephemeral`, the VMI is left alone and the Machine is handed back for replacement right away, instead of burning `retryLimit × timeoutSeconds` on a node that cannot rejoin. The check is deliberately conservative (any such volume, not just the boot disk). `emptyDisk`/`hostDisk` are not inspected — an emptyDisk is scratch space that is expected to start empty — and the docs say they must not hold node state. The persistent-storage requirement is documented, including the note that most shipped templates use containerDisk and therefore degrade to plain replacement.

Related: only failed preconditions hand the Machine over now. Transient errors (e.g. infra cluster unreachable) are returned and retried with backoff, so a blip no longer replaces a Machine.

controllers/kubevirtremediation_controller.go
changjjjjjjj: Done. Reconcile now returns without restarting, escalating or writing status while `Cluster.spec.paused` is set or the KubevirtRemediation / the Machine carries the paused annotation (the MachineHealthCheck controller skips paused Machines the same way). For resuming, `SetupWithManager` adds `ResourceNotPaused`, a Cluster watch with `ClusterUnpaused`, and a Machine watch.

For the stale-request case you describe: since the MachineHealthCheck is paused for the same time, `HealthCheckSucceeded` can still say `False` for a node that recovered during the pause. To avoid acting on that, the controller stays idle for one minute after the Machine's `Paused` condition turns `False`, giving the MachineHealthCheck time to re-evaluate and delete the request; and once a boot window has expired it never retries or escalates while `HealthCheckSucceeded` is `True`. Both are covered by unit tests (pause ×3, no escalation while paused, resume, grace period).

controllers/kubevirtremediation_controller_test.go
changjjjjjjj: Added, in the same commit:

**e2e** (`Context("with external remediation")`, helpers in `e2e/remediation_test.go`). The MachineHealthCheck of these specs only looks at a node condition owned by the tests (like a node-problem-detector condition, it survives the restart), so a spec decides when the node is unhealthy and when it has recovered, independently of boot times.
- *Recovery + pause during a pending retry* (persistent storage): fault → first restart → pause the Cluster → the boot window expires with the fault still set, and neither the VMI UID nor `retryCount` move → Node Ready again → clear the fault → unpause → the MachineHealthCheck removes the request and the node is not restarted a second time. Asserts a new VMI UID and boot ID, and unchanged Machine, VM, PVC and Node UIDs plus the guest's machine-id.
- *Retry exhaustion → replacement, on external infrastructure with persistent storage*: the persistent-storage template with `infraClusterSecretRef` injected, so the VMI delete really goes through the external kubeconfig (the spec checks the ref on the KubevirtMachine first).
- *No persistent storage* (default template): `Failed` with `retryCount` 0, then replacement.
- An inner `AfterEach` unpauses the Cluster so a failing spec cannot leave an undeletable cluster behind. The kubevirtci overlay turns the flag on. Heads-up: the suite now takes a bit over an hour, which is past Ginkgo's default suite timeout, so `hack/run-e2e.sh` sets `-ginkgo.timeout=2h` (the prow job allows 3h).

**Manager startup** (`main_test.go`, envtest, `make test-integration`, new CI job `envtest`; skipped by `make test` when the envtest binaries are absent): without the remediation CRDs and the flag off, the manager keeps running past the cache-sync timeout and the existing controllers reconcile; with the flag on and no CRDs the setup fails with a no-match error; with CRDs and the flag on it additionally verifies the watches (a request is ignored while its Cluster is paused and processed on unpause). `setupReconcilers` now returns its error instead of exiting, to make that testable.

Two behavior changes since my earlier replies, found while writing these: the post-unpause grace period looks at the `Paused` conditions of both the Cluster and the Machine (the Machine's lags for Machines not owned by a MachineSet), and a Machine is never handed back while `HealthCheckSucceeded` is still `True` (the MachineSet would drop the condition).

